PR #127, opened 2019-04-30
added "startedat" field to tests-finished event

bacherfl:


jetzlstorfer:
no typo. attribte has the meaning of "started at", lowercase writing since all cloud events have lowercasing

PR #135, opened 2019-05-14
Clarify prerequisites

jetzlstorfer:
- added comment on prerequisite of use case


jetzlstorfer:
Updated description. Could you please take another look @dirkwall ? Thanks!

PR #227, opened 2019-07-19
#548 Use API component instead of control

agrimmer:


johannes-b:
Due to the new keptn release, a new section is provided on the live website (Release 0.4.0). The file changes point to this section (0.4.0 folder in content/docs) that cannot be modified. 
I created a separate PR (#240), which changes the files in the develop folder.

PR #340, opened 2019-10-23
Update develop based on changes in master

johannes-b:


johannes-b:
This PR is not required since no changes in content made.

PR #352, opened 2019-11-11
Edits to the Setup GKE Content and the First Fold of the Homepage

leticiamendonca:
Tested the content in the Setup GKE file, and edited both the options - Console and CLI. Also edited the content that appears on the first fold of the Keptn homepage.


johannes-b:
Hi Leticia, 

many thanks for your PR. 
As you can see on the new keptn.sh landing page, a couple of issues addressed in your PR have been resolved in a new manner. 
This is the reason why the changes in this PR are not applicable and I would like to close it.

PR #391, opened 2019-12-19
remove outdated information about health checks

jetzlstorfer:


christian-kreuzberger-dtx:
Health check is still executed automatically:

https://github.com/keptn/keptn/blob/0.6.0.beta2/jmeter-service/main.go#L70-L88

so this information is not outdated.

PR #392, opened 2019-12-19
#1246 Setup monitoring for Dynatrace, fixed spelling issues 

johannes-b:
- Update Keptn CLI command: `keptn configure monitoring`
- Update installation instructions for Dynatrace monitoring to reflect the current implementation
- Fixed some spelling issues


content/docs/0.6.0/reference/monitoring/dynatrace/index.md
```diff
@@ -52,55 +35,35 @@ Please make sure to have the following tool(s) installed:
...
 
     ```console
-    ./deployDynatraceOnOpenshift.sh
+    keptn configure monitoring dynatrace
     ```
 
-When this script is finished, the Dynatrace OneAgent and the *dynatrace-service* are deployed in your cluster. Execute the following commands to verify the deployment of the OneAgent as well as of the *dynatrace-service*.
+**Verify deployment in your cluster**
+
+When keptn configure monitoring is finished, the Dynatrace OneAgent is deployed in your cluster. Execute the following commands to verify the deployment of the OneAgent as well as of the *dynatrace-service*:
```
agrimmer: Use source formatting for `keptn configure monitoring`?

content/docs/0.6.0/reference/monitoring/dynatrace/index.md
```diff
@@ -133,44 +117,31 @@ kubectl delete pods --all --namespace=sockshop-staging
 kubectl delete pods --all --namespace=sockshop-production
 ```
 
-**Note 2:** If the nodes in your cluster run on *Container-Optimized OS (cos)* (default for GKE), the Dynatrace OneAgent might not work properly, and another step is necessary. To verify that the OneAgent does not work properly, the output of `kubectl get pods -n dynatrace` might look as follows:
+**Note 3:** If the nodes in your cluster run on *Container-Optimized OS (cos)* (default for GKE), the Dynatrace OneAgent might not work properly, and another step is necessary. To verify that the OneAgent does not work properly, the output of `kubectl get pods -n dynatrace` might look as follows:
```
agrimmer: Describe what has to be done when the OneAgent is not working properly.

content/docs/0.6.0/reference/monitoring/dynatrace/index.md
```diff
@@ -133,44 +117,31 @@ kubectl delete pods --all --namespace=sockshop-staging
...
 oneagent-k7jn6                                 0/1     CrashLoopBackOff   6          8m15s
 ```
-This means that after the initial setup with `deployDynatrace.sh`, which is a step below, the `cr.yml` has to be edited and applied again.  You can do that by editing the already downloaded `cr.yml` in `../manifests/dynatrace/gen` and set the environemnt variable as follows:
 
-```yaml
-  env:
-  - name: ONEAGENT_ENABLE_VOLUME_STORAGE
-    value: "true"
-```
+1. This means that after the initial setup with `deployDynatrace.sh`, which is a step below, the `cr.yml` has to be edited and applied again.  You can do that by editing the already downloaded `cr.yml` in `../manifests/dynatrace/gen` and set the environment variable as follows:
```
agrimmer: This needs to be adapted because `deployDynatrace.sh` does not exist anymore.

content/docs/0.6.0/reference/monitoring/dynatrace/index.md
johannes-b: Is explained in the following steps (1, 2, and 3).

content/docs/0.6.0/reference/monitoring/dynatrace/index.md
johannes-b: Thx, the command is now properly formatted.

content/docs/0.6.0/reference/monitoring/dynatrace/index.md
johannes-b: Thanks for pointing that out. I updated the instructions to reflect the current approach for this situation.

PR #406, opened 2020-01-10
Added promotion header with GKE partner link

jetzlstorfer:
link to GCP credits if you apply via Keptn added on
- landing page
- quickstart with GKE
- install instructions with GKE


jetzlstorfer:
CSS updates seem not to go through. Links should be white instead of gray. @martinhellwagner can you help me out here? thanks!

layouts/index.html
```diff
@@ -26,6 +26,23 @@
...
+{{/* Promo */}}
+<div class="promo section-primary">
+  <div class="container pt-1 pb-1 text-center">
+    <div class="row pt-md-5 pb-md-5">
+      <div class="col-12 col-md-12 col-lg-12">
+        {{ with .Site.GetPage "section" "home/promo" }}
+          {{ .Content }}
+        {{ end }}
+        <br>
+        <a class="button button-secondary" href="http://bit.ly/keptnongke" target="_blank">Sign up and use Keptn in GKE</a>
```
christian-kreuzberger-dtx: I'm not the one to decide, but I have concerns in relying on a third party like *bit.ly* for a promotional link instead of a direct link.

jetzlstorfer: With the direct link we would not know how many people are actually making use of the promotion, therefore we decided to use bitly

PR #415, opened 2020-01-20
minor fixes for qg-standalone use case

bacherfl:


christian-kreuzberger-dtx:
@johannes-b please also take a look

PR #419, opened 2020-01-21
Patch/rename dt permission

Dynatrace-Adam-Gardner:
Patch docs to update dynatrace permission to "Read log content".
Also update image to reflect this.


johannes-b:
Hi Adam, thanks for the PR!

PR #438, opened 2020-02-07
Added develop folder as a preparation for 0.6.1

christian-kreuzberger-dtx:


johannes-b:
Should we conduct all doc changes for 0.6.1 in the official 0.6.x documentation? I don't see a need to provide separate documentation for 0.6.1.

christian-kreuzberger-dtx:
Doc "fixes" will be done in 0.6.0, anything that is new (e.g., new Keptn's Bridge, behavioural changes of lighthouse, etc ...) should be done in the new develop folder. I've left the folder empty on purpose for now, we will copy the original files when we need it.

PR #451, opened 2020-02-28
Updated quickstart guide

christian-kreuzberger-dtx:
- Added `kubectl` as a requirement
- Added a quick verification step using `keptn status`


christian-kreuzberger-dtx:
That's a valid point, I'll move it.

PR #454, opened 2020-03-06
re-added runbook automation tutorial

bacherfl:
The documentation of that use case now points to the servicenow-service github repo


bacherfl:
the release tag is now available

PR #470, opened 2020-03-18
add tutorials in menu

jetzlstorfer:
Add a link to the tutorials to the main menu.


jetzlstorfer:
I can see it now. tested it locally with `hugo server` and looked fine. let me investigate this.

jetzlstorfer:
hugo is creating a dedicated page for each menu entry by default. I have to take a look how to overwrite this to reuse already existing pages (in my case the tutorials). 
I'll put this on hold for now and will take a look later.

PR #524, opened 2020-04-28
Set different color for develop branch (#523) 

johannes-b:


johannes-b:
Has someone Hugo "extended" installed and can generate the SCSS files in this PR?

PR #542, opened 2020-05-14
Updated K8s support matrix for Keptn 0.7

johannes-b:


content/docs/develop/operate/k8s_support/index.md
```diff
@@ -20,18 +20,25 @@ Keptn versions are expressed as `x.y.z`, where `x` is the major version, `y` is
...
-**Remarks:**
+**Remarks for Keptn 0.7**
 
 * (1): AKS with K8s version before 1.15 might fail [#1429](https://github.com/keptn/keptn/issues/1429), due to a known AKS issue: [#69262](https://github.com/kubernetes/kubernetes/issues/69262)
+* (3): GKE does not provide K8s 1.16 clusters (checked on: 13th May, 2020)
+
+**Remarks for Keptn 0.6**
+
+* (1): AKS with K8s version before 1.15 might fail [#1429](https://github.com/keptn/keptn/issues/1429), due to a known AKS issue: [#69262](https://github.com/kubernetes/kubernetes/issues/69262)
+* (2): EKS and GKE do not provide K8s 1.16 clusters (checked on: 9th April, 2020)
```
christian-kreuzberger-dtx, line 37: not sure if we should do the split down here, I think it's enough to have the remarks (1), (2) and (3) (in that order), without the title of the version

johannes-b: Removed remark (3) since GKE 1.16 is available. Thus, this version needs to be supported by Keptn 0.7.0. 

This also makes the split of the remarks obsolete.

christian-kreuzberger-dtx:
I've tested K3s and added integration tests for the quality-gates test to the keptn/keptn repo.

I believe we can add K3s to this list as well (versions 1.16, 1.17 and 1.18).

In addition, I think we should make this table a little bit easier to read, especially when multiple versions are compatible (Minikube is such a case). 
Instead of explicitly stating every version that we are testing with, how about just providing a version range?

johannes-b:
* I added K3s to the support matrix. 

* Agree, the table becomes unreadable. I compressed it by providing ranges instead of listing all supported versions: 
image
This makes it more reader-friendly.

PR #553, opened 2020-05-28
Update of required Dynatrace API token

johannes-b:
Remove permissions from required Dynatrace API token (RUM javascript tag not needed)


christian-kreuzberger-dtx:
I believe we need this token for real user monitoring / app detection, as described in issue https://github.com/keptn/keptn/issues/1024 and implemented in PR https://github.com/keptn-contrib/dynatrace-service/pull/105 

What's the reason for removing the token from the list now?

johannes-b:
This update is necessary since: 
- **Access logs** has been renamed to **Read log content**. 
- **Configure maintenance windows** is marked as deprecated.

That is true, for the feature https://github.com/keptn/keptn/issues/1024 the API token needs: **Real user monitoring JavaScript tag management**. However, this feature is not released and we should restrict the API token to the least minimum set of permissions. 

I have added a comment https://github.com/keptn/keptn/issues/1024#issuecomment-643984919 to the issue that an update of the Dynatrace API token is needed when using the implemented feature.

PR #557, opened 2020-06-04
#1870 Add replicaCount in values

agrimmer:
Potentially fixes https://github.com/keptn/keptn/issues/1870


johannes-b:
I have to withdraw my approval. We must fix it here as well: https://keptn.sh/docs/develop/manage/service/

PR #597, opened 2020-07-24
Re-work installation and split into options

agrimmer:


content/docs/0.7.x/operate/install/index.md
```diff
@@ -219,288 +227,224 @@ Every Keptn release provides binaries for the Keptn CLI. These binaries are avai
...
+    KEPTN_API_ENDPOINT=http://<HOST>/api
+    ```
 
-<details><summary>Keptn API endpoint when exposing via **LoadBalancer**</summary>
-<p>
+:warning: **Warning:** If you do not setup TLS encryption, all your traffic to and from the Keptn endpoint is not encrypted.
 
-* If you are using a **LoadBalancer** to expose Keptn, get the EXTERNAL-IP of the `api-gateway-ngix` using the next command. Consequently, the Keptn API endpoint is: `http://<ENDPOINT_OF_API_GATEWAY>/api`
+### Option 4: Access Keptn via a Port-forward
+This option exposes Keptn on a *cluster-internal* IP .
```
johannes-b, line 417: Maybe we could write the sentence like that: _This option does not expose Keptn, but rather allows accessing Keptn on its *cluster-interanl* IP._

To make it clear that this is not exposing Keptn.

content/docs/0.7.x/operate/install/index.md
```diff
@@ -219,288 +227,224 @@ Every Keptn release provides binaries for the Keptn CLI. These binaries are avai
...
-
-  ```console
-  oc get route api -n keptn
-  ```
   
-  ```
-  NAME        HOST/PORT                       PATH      SERVICES            PORT      TERMINATION   WILDCARD
-  api         api-keptn.x.x.x.xip.io                    api-gateway-nginx   http      edge/None     None
-  ```
+## Authenticate Keptn CLI
```
johannes-b, line 439: Please do a find `#5-authenticate-keptn-cli` and replace with `authenticate-keptn-cli` in the entire docu.

PR #663, opened 2020-10-13
#2429 Create page with details about the integrations page in Bridge

ermin-muratovic:
Create page with details about the integrations page in Bridge (GDPR info)


johannes-b (approved):
lgtm

PR #671, opened 2020-10-28
#666 updated screenshot and description

### LINKED ISSUE #666: Update Dynatrace Screen for creating secrets
In https://keptn.sh/docs/0.7.x/monitoring/dynatrace/install/ we are showing a screenshot of the (soon to be) outdated API token screen.

This screen will soon have two columns, separated for API v1 and API v2. This is what the new screen design looks like:
image

On the API V1 side, things stay the same as with the old screenshot.
On the API V2 side, we need: READ metrics, READ entities, and in the future INGEST Metrics (e.g., OpenTelemetry, Prometheus Ingest)

warber:
Closes #666


johannes-b:
Hi @warber,
thanks for the PR. Looks good so far. 
But we also need to update it in the `0.7.x` folder. Can you please do that update?

warber:
> But we also need to update it in the `0.7.x` folder. Can you please do that update?
sure, done

johannes-b (approved):
thanks! lgtm

PR #678, opened 2020-11-04
#2606 Added doc for installation type env var of bridge

bacherfl:
Related to [#2609](https://github.com/keptn/keptn/issues/2609)


johannes-b (approved):
lgtm

PR #679, opened 2020-11-10
Use create instead of replace when creating Secret or ConfigMap

johannes-b:
Based on user feedback, the command: 
`kubectl create configmap -n keptn ingress-config --from-literal=ingress_hostname_suffix=${INGRESS_HOSTNAME} --from-literal=ingress_port=${INGRESS_PORT} --from-literal=ingress_protocol=${INGRESS_PROTOCOL} --from-literal=istio_gateway=${ISTIO_GATEWAY} -oyaml --dry-run | kubectl replace -f -` does not work on a fresh cluster, since the configmap is not available. 

I fixed it by changing it to `create` instead of `replace`.

### DIFF
--- a/content/docs/0.7.x/continuous_delivery/expose_services/index.md
+++ b/content/docs/0.7.x/continuous_delivery/expose_services/index.md
@@ -64,7 +64,7 @@ kubectl apply -f gateway-manifest.yaml
       **Note:** In the above example, `xip.io` is used as wildcard DNS for the IP address.
 
     ```console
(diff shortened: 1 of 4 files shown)


christian-kreuzberger-dtx:
Can not confirm, I just tried
```
$ kubectl -n keptn create secret generic dynatrace --from-literal="DT_API_TOKEN=$DT_API_TOKEN" --from-literal="DT_TENANT=$DT_TENANT" --from-literal="KEPTN_API_URL=$KEPTN_API_URL" --from-literal="KEPTN_API_TOKEN=$KEPTN_API_TOKEN" -oyaml --dry-run | kubectl create -f -
W1124 09:43:59.475772  107674 helpers.go:553] --dry-run is deprecated and can be replaced with --dry-run=client.
secret/dynatrace created
```
on a fresh cluster and it worked.

However, if the secret already exists, `kubectl create -f` does not work:
```
Error from server (AlreadyExists): error when creating "STDIN": secrets "dynatrace" already exists
```

In addition, using `replace` does not work if the secret does not exist (on a fresh cluster):
```
Error from server (NotFound): error when replacing "STDIN": secrets "dynatrace" not found
```

christian-kreuzberger-dtx:
Closing it, as I believe it's not needed.

Issue #673, opened 2020-10-30
Update documentation about Keptn architecture and provide recommendation for cluster size


johannes-b:
**Architecture:**

- The description of the architecture of Keptn (as shown here: https://keptn.sh/docs/concepts/architecture/) needs to be updated since components are gone or have been replaced. Besides, it must be explained what the execution plane is and how it interacts with the control plane. 

- The architecture diagram should be in a higher resolution as it is right now. 

**Cluster size:**

- Add a recommendation for the cluster size Keptn is running on:
  - when installing control-plane only
  - when installing control-plane + execution-plane
  - plus rough estimation about requirements (or just explanation) for apps. 

- Add the sizing table to: https://keptn.sh/docs/0.8.x/operate/k8s_support/ (with a reference before installing Keptn)

--> The automation of creating this table will be implemented in: https://github.com/keptn/keptn/issues/1512

PR #683, opened 2020-11-12
#673 Provide overview of resource consumption

johannes-b:
Addresses one task of #673


christian-kreuzberger-dtx:
Can we automate this? Else we have another step that we need to do every time we create a release.

christian-kreuzberger-dtx (approved):
LGTM

PR #689, opened 2020-11-24
Explain how to access Keptn Bridge in Quickstart

johannes-b:
This PR explains how to access the Keptn Bridge after installing Keptn.

### DIFF
--- a/content/docs/quickstart/_index.md
+++ b/content/docs/quickstart/_index.md
@@ -116,7 +116,7 @@ Please make sure you have `kubectl` installed (see [kubernetes.io/docs/tasks/too
     keptn version
     ```
 
-#### 2.2 Install Keptn and authenticate Keptn CLI
+### 3. Install Keptn and authenticate Keptn CLI
 
 * By executing the [keptn install](../0.7.x/reference/cli/commands/keptn_install) command as shown next, Keptn will be installed on your Kuberentes cluster supporting all continuous delivery use cases (including quality gates and automated operations):
 
@@ -144,7 +144,7 @@ Please make sure you have `kubectl` installed (see [kubernetes.io/docs/tasks/too
 
 :rocket: Keptn is now ready to be used.
 
-### 3. Access Keptn Bridge
+**Access Keptn Bridge**
 
 By default, basic authentication is in place that protects the Keptn Bridge. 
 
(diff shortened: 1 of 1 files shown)


content/docs/quickstart/_index.md
```diff
@@ -116,7 +116,7 @@ Please make sure you have `kubectl` installed (see [kubernetes.io/docs/tasks/too
     keptn version
     ```
 
-#### 3. Install Keptn and authenticate Keptn CLI
+#### 2.2 Install Keptn and authenticate Keptn CLI
```
christian-kreuzberger-dtx, line 120: If we change the headers, we will break all existing links to this page. I advise against changing this to 2.2

content/docs/quickstart/_index.md
```diff
@@ -144,6 +144,20 @@ Please make sure you have `kubectl` installed (see [kubernetes.io/docs/tasks/too
...
+
+By default, basic authentication is in place that protects the Keptn Bridge. 
+
+* To get the `user` and `password`, execute the command:
+
+    ```console
+    keptn configure bridge --output
+    ```
+
+* Depending on the chosen way of exposing Keptn (please see [here](../0.7.x/operate/install/#install-keptn)), you need to derive the endpoint Keptn is running on. If you follow this quickstart, the endpoint is: `http://localhost:8080`
```
christian-kreuzberger-dtx, line 157: We should add that it is posible to use `keptn status` to figure out the current API endpoint address:
```console
$ keptn status
Starting to authenticate
Successfully authenticated against the Keptn cluster https://keptn.example.com/api
Using a file-based storage for the key because the password-store seems to be not set up.
CLI is authenticated against the Keptn cluster https://keptn.example.com/api
```
Consequently, one would need to access https://keptn.example.com

content/docs/quickstart/_index.md
johannes-b: Valid point - changed it back.

content/docs/quickstart/_index.md
johannes-b: Good idea. I added the `keptn status` as a mean for getting the endpoint.

christian-kreuzberger-dtx (approved):
LGTM

PR #727, opened 2021-02-03
Removed outdated information regarding installer

christian-kreuzberger-dtx:

Since Keptn 0.7.x we have switched to a helm based installation method, therefore we don't have an installer anymore, and we don't need to provide troubleshooting details for it.


Kirdock (approved):
LGTM

PR #750, opened 2021-02-18
Fix problem with teaser banner and header

martinhellwagner:


jetzlstorfer:
Thanks for the PR - reviewing right now :)

martinhellwagner:
@jetzlstorfer 

Basically, you just have to remove or comment out `{{ partial "announcement.html" . }}` in order for the header to be pushed up correctly. When you are ready to show the teaser banner again, add or comment in the partial again. 🙂

jetzlstorfer:
works as advertised :) 

but there is an animation each time the page is reloaded. can we remove this animation? for the first 0.5 seconds or so the teaser and the menu are overlapping. if there is no teaser, it looks good, no animation there.

jetzlstorfer:
@martinhellwagner I fixed it by making the "pushed-down version" the default one and doing the "move" of the navigation one the announcement is not present. This way it feels a bit smoother. Could you please have a look if I did it correctly?

martinhellwagner:
@jetzlstorfer 

Looks good! You could also check for `if (!teaser && teaser == null)` for an extra layer of checks, but I guess in our case, the `null` check should be fine as well.

jetzlstorfer:
thanks for the input @martinhellwagner ! I'll have it reviewed and we'll merge it

jetzlstorfer (approved):
lgtm

PR #753, opened 2021-02-19
Update istio ingress for api-service

christian-kreuzberger-dtx:

During setup of 0.8.0-dev I noticed that the istio ingress for api-gateway-nginx is not working anymore with istio 1.8.

As Istio 1.8+ is our current istio target version, I've updated it here accordingly.


johannes-b (approved):
lgtm

PR #754, opened 2021-02-19
add starttime to videos

jetzlstorfer:
Adding timestamp to video so they start directly with presentation and skip the intro words


johannes-b (approved):
lgtm

PR #763, opened 2021-03-01
Replace xip.io with nip.io

christian-kreuzberger-dtx:

While going through the setup of helm-service, I noticed that we are still referring to xip.io, which is highly unreliable and causes problems.
This PR replaces occurrences of xip.io with nip.io (for 0.7.x and 0.8.x).


johannes-b (approved):
lgtm

PR #764, opened 2021-03-03
Fix distributor URLs for running in execution plane

christian-kreuzberger-dtx:

Related to https://github.com/keptn/keptn/issues/3418

While the distributor itself was updated to other URLs, we did not update docs accordingly. This PR fixes it.


johannes-b (approved):
lgtm

PR #765, opened 2021-03-04
Update get.keptn.sh quickstart instructions

christian-kreuzberger-dtx:

This PR updates instructions when downloading the CLI via get.keptn.sh.

* we don't use `sudo -E bash` anymore, `bash` is enough
* old versions of Keptn can still be downloaded, but `KEPTN_VERSION=x.y.z` needs to be specified


johannes-b (approved):
lgtm 
(great that we support the version as a parameter now)

PR #801, opened 2021-04-28
Disable relative URLs. fixes #794

### LINKED ISSUE #794: Documentation 404 pages could be enhanced
When navigating to a page that returns 404, no assets (images, css, etc. are loaded) and therefore the page looks barebone.

E.g.
https://keptn.sh/docs/0.8.x/reference/cli/commands/asdfasdf
https://keptn.sh/community/asdfasdf

image

This seems to be a problem with path locations other than the first one, as https://keptn.sh/asdf delivers a proper 404 page.

anthu:
As outlined in #794 the 404 pages have broken CSS.

Hugo is assuming the 404 page will appear under `/404/index.html`. But as the web server is not redirecting to 404 but just rewriting the request these paths are being messed up.

Switching off relative linking should not only resolve this issue but also avoid links like

```
<link rel="stylesheet" href="../../../../../../js/libs/doc/lightbox/css/lightbox.min.min.1bbb47ec548f1c47e276ce27971cdc27f8c803ee4bb96f064c72c7448f7c0f12.css">
```

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -2,7 +2,7 @@
 languageCode = "en-us"
 title = "keptn | Cloud-native application life-cycle orchestration"
 theme = "hugo-serif-theme"
-relativeUrls = true
+relativeUrls = false
 enableEmoji = true
 
 ignoreFiles = [


jetzlstorfer:
Hi @anthu - thanks for the PR! 
Looks like a great fix. 

I verified and broken links such as the one I made up here https://deploy-preview-801--keptn.netlify.app/community111/ now have a correct styling.

jetzlstorfer (approved):
LGTM

Issue #823, opened 2021-05-25
Revamp quickstart section


jetzlstorfer:
Currently, the **quickstart** section gives a lot of different options on how to install Keptn. 

**Desired solution:**
The quickstart should allow going from "0 to bridge" within only 5 minutes. Therefore, I suggest providing a local option such as k3d for the quick start guide and give a clear set of instructions. If you follow them, you'll have a successful Ketpn installation including access to the bridge within < 5 minutes.

**Proposed steps**

- [ ] Install k3d
- [ ] Install Istio 
- [ ] Download Keptn CLI 
- [ ] Install Keptn in K3d
- [ ] Configure Istio 
- [ ] Authenticate Keptn
- [ ] Create project
- [ ] Create service
- [ ] Open Bridge

jetzlstorfer:
For the setup of K3d and Istio, I'll follow this guide https://jetzlstorfer.medium.com/running-k3d-and-istio-locally-32adc5c41a63

PR #826, opened 2021-05-27
New quickstart guide to run Keptn locally

jetzlstorfer:
Addresses #823 

### DIFF
--- a/content/docs/quickstart/_index.md
+++ b/content/docs/quickstart/_index.md
@@ -7,6 +7,41 @@ weight: 1
 hidechildren: true # this flag hides all sub pages in the sidebar-multicard.html
 ---
 
+We'll run Keptn on a local [k3d](https://k3d.io) cluster.
+
+## Prerequisites
+- [Docker](https://docker.com/)
+- [k3d](https://k3d.io) (find [installation instructions](https://k3d.io/#install-script) or run<br> `wget -q -O - https://raw.githubusercontent.com/rancher/k3d/main/install.sh | bash` )
+- [kubectl](https://kubernetes.io/docs/tasks/tools/#kubectl)
+- [helm](https://helm.sh/) when running the demos in step 4
+
+## Install Keptn
+
+1. Create local k3d cluster
+
+    ```
+    k3d cluster create mykeptn -p "8082:80@agent[0]" -p "8443:443@agent[0]" --k3s-server-arg '--no-deploy=traefik' --agents 1
+    ```
+
+1. Get Keptn CLI
+
+    ```
+    curl -sL https://get.keptn.sh | bash
+    ```
+
+1. Install Keptn control plane + execution plane for continuous delivery
+    ```
+    curl -sL QUICKSTART.SH | bash 
+    ```
+    If you want to install Keptn yourself without following the quickstart guide, have a look at the [install options](../).
+
+1. (Optional but recommended) Create a demo project with multi-stage pipeline + SLO-based quality gates
+    ```
+    curl GET_DEMO.sh | bash
+    ```
+
+
+
 ### 1. Create Kubernetes cluster
 
 Select one of the following options:
--- a/content/docs/quickstart/_index.md
+++ b/content/docs/quickstart/_index.md
@@ -1,26 +1,30 @@
 ---
 title: Quick Start
-description: Learn how to get Keptn running in five minutes.
+description: Learn how to get Keptn running in five minutes. We'll run Keptn on a local k3d cluster.
 icon: concepts
 layout: quickstart
 weight: 1
 hidechildren: true # this flag hides all sub pages in the sidebar-multicard.html
 ---
 
-We'll run Keptn on a local [k3d](https://k3d.io) cluster.
 
 ## Prerequisites
 - [Docker](https://docker.com/)
-- [k3d](https://k3d.io) (find [installation instructions](https://k3d.io/#install-script) or run<br> `wget -q -O - https://raw.githubusercontent.com/rancher/k3d/main/install.sh | bash` )
 - [kubectl](https://kubernetes.io/docs/tasks/tools/#kubectl)
-- [helm](https://helm.sh/) when running the demos in step 4
+- [helm](https://helm.sh/) when running the demos in step 5
 
 ## Install Keptn
 
 1. Create local k3d cluster
 
+    First, install [k3d](https://k3d.io) if not already present on your machine:
     ```
-    k3d cluster create mykeptn -p "8082:80@agent[0]" -p "8443:443@agent[0]" --k3s-server-arg '--no-deploy=traefik' --agents 1
+    curl -s https://raw.githubusercontent.com/rancher/k3d/main/install.sh | TAG=v4.4.4 bash
+    ```
+
+    Now, let's start a cluster for Keptn!
+    ```
+    k3d cluster create mykeptn -p "8082:80@agent[0]" --k3s-server-arg '--no-deploy=traefik' --agents 1
     ```
 
 1. Get Keptn CLI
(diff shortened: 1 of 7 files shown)


jetzlstorfer:
Preview: https://deploy-preview-826--keptn.netlify.app/docs/quickstart/

christian-kreuzberger-dtx:
# my two cents

To me, this feels like too much magic is hidden in Bash Scripts.

image

**General Question**: Where will does bash scripts be hosted? Who will maintain those bash scripts? Who will make sure those bash scripts work with every possible system (Linux vs. Windows vs. MacOS, WSL, Git Bash, ...)? I really advise against this approach. But I have alternative ideas, which I've outlined below.

**ad 1.**: While I heavily prefer to use K3s (I'm on Ubuntu btw), I understand that not everybody can use K3s (Windows? Mac?), correct? Can we let the user decide (K3d, K3s, GKE, Bring your own K8s cluster) in the beginning, but make *K3d* the preferred and automatically expanded option?
**Side Note**: We do not have an integration test with K3d, so whatever happens there, is absolutely untested and we have no way of knowing if it breaks. We have a test for K3s though.

**ad 3.**: Please don't use a bash script for **just installing** Keptn. This will be very hard to maintain and for the end-user this is pure magic. IMO the user really just needs to type `keptn install` (or `keptn install --use-case=continuous-delivery`). If we need to have Istio installed and pre-configured (e.g., for a fully functional CD setup), then let's make a separate step called *install_and_configure_istio.sh*. That way the user knows the responsibilities of each step, and what to expect.

**ad 4.**: This is a great idea. I think this should be functionality that is accessible via the CLI, e.g., `keptn create project foobar --import=https://github.com/keptn/quality-gate-example-pipeline.git@0.8.3`. I know this is not something we can do right now, but I think the benefit is there for quickstart and for Keptn end-users. So let's make sure this step gets added as a story to Keptn CLI.

jetzlstorfer:
Hi @christian-kreuzberger-dtx thanks for your comments!

The main purpose of this is to provide a way to get started with Keptn quickly and provide value within a couple of minutes. Right now, the quickstart guide is too complicated and gives too many options. The idea is to provide ONE way to do it that should be generally applicable. In my research, I found that K3d is the best solution here.
In general, the user can always decide where to install Keptn, as outlined here: https://keptn.sh/docs/0.8.x/operate/install/ but for the quickstart it should be QUICK and not going through multiple options on how to create a cluster.

One of the strengths of Keptn is its capability to integrate with and orchestrate other tools. The drawback is that if the user needs to set up all these tools himself, this takes quite a while, but the user will learn a lot. Therefore we have provided the tutorials to guide the user on this journey. For a more generic approach, users will follow the documentation. But for a quickstart, we can give an opinionated way to do it and this is the purpose of this PR.

**ad 1:** It would be desirable to have a maintained quickstart guide for each platform, but I believe this needs some more resources on the engineering side. Currently, I would go for 1 solution for the quickstart (k3d), and for other installation options provide a link to the general documentation. K3s is not a good option as it excludes MacOS and Windows users. A follow up issue could be to provide integration tests also for K3d.

**ad 3:** The script does not only install Keptn but also sets up Istio, disables the login for the bridge (again, it is a quickstart that should give you access to a running installation in < 3 minutes) and opens the Bridge for the user to be ready to be explored. For me it is ok to separate these two parts in two commands: 1. install Keptn and 2. Install and configure Istio + disabling auth and opening the bridge

**ad 4:** It would be great to have this possibility via the CLI. But even if we create a project including services+tests+SLI+SLO via the CLI, the user still has to 1. Install Prometheus, 2. Install Prometheus-Service, (3. Install Prometheus-sli-service which is about to get merged into the Prometheus-service anyway), 4. Configure Prometheus, 5. Trigger one or more delivery sequences 
Therefore we need some kind of scripting magic that does this for the user (we can explain everything that is going on with printing it to the stdout while the script is executing).

Summing up, I still advocate for providing a simplified quickstart on a local machine via scripts (as long as the CLI can not serve this purpose).

content/docs/quickstart/_index.md
```diff
@@ -7,6 +7,41 @@ weight: 1
...
+- [k3d](https://k3d.io) (find [installation instructions](https://k3d.io/#install-script) or run<br> `wget -q -O - https://raw.githubusercontent.com/rancher/k3d/main/install.sh | bash` )
+- [kubectl](https://kubernetes.io/docs/tasks/tools/#kubectl)
+- [helm](https://helm.sh/) when running the demos in step 4
+
+## Install Keptn
+
+1. Create local k3d cluster
+
+    ```
+    k3d cluster create mykeptn -p "8082:80@agent[0]" -p "8443:443@agent[0]" --k3s-server-arg '--no-deploy=traefik' --agents 1
```
christian-kreuzberger-dtx, line 23: Please include the installation instructions for k3d here, e.g.:
```
wget -q -O - https://raw.githubusercontent.com/rancher/k3d/main/install.sh | bash
k3d cluster create mykeptn -p "8082:80@agent[0]" -p "8443:443@agent[0]" --k3s-server-arg '--no-deploy=traefik' --agents 1
```
Also, please use a fixed version of k3d.

content/docs/quickstart/quickstart.sh
```diff
@@ -0,0 +1,123 @@
...
+echo "Creating Ingress config for Keptn"
+kubectl create configmap -n keptn ingress-config --from-literal=ingress_hostname_suffix=$(kubectl -n keptn get ingress api-keptn-ingress -ojsonpath='{.spec.rules[0].host}') --from-literal=ingress_port=$INGRESS_PORT --from-literal=ingress_protocol=http --from-literal=istio_gateway=public-gateway.istio-system -oyaml --dry-run=client | kubectl apply -f -
+
+# Restart helm service
+echo "Restarting helm-service to load new settings"
+#kubectl delete pod -n keptn -lapp.kubernetes.io/name=helm-service
+
+echo "Authenticating Keptn CLI against Keptn installation"
+echo "keptn auth --endpoint=http://$INGRESS_IP.nip.io:$INGRESS_PORT --api-token=*****"
+keptn auth --endpoint=http://$INGRESS_IP.nip.io:$INGRESS_PORT --api-token=$(kubectl get secret keptn-api-token -n keptn -ojsonpath={.data.keptn-api-token} | base64 --decode)
```
christian-kreuzberger-dtx, line 95: I am able to just use `keptn auth` without parameters. `keptn` CLI is able to fetch API connection details via `kubectl` in most cases!

content/docs/quickstart/quickstart.sh
```diff
@@ -0,0 +1,123 @@
+#!/bin/bash
+set -e
+
+# istio settings
+ISTIO_VERSION=1.10.0
+INGRESS_IP=127.0.0.1
+INGRESS_PORT=8082
```
christian-kreuzberger-dtx, line 7: This might not always work.

For me on Linux I did not set those variables, and used the installation script from the tutorials
```
curl -o configure-istio.sh https://raw.githubusercontent.com/keptn/examples/release-0.8.3/istio-configuration/configure-istio.sh
```

where we fetch the ingress IP (and do not need a port).

For me on Linux it lead to

```
http://10.XXX.YYY.2.nip.io
```

content/docs/quickstart/_index.md
jetzlstorfer: done

content/docs/quickstart/quickstart.sh
jetzlstorfer: not for me as I am using not the standard port 80 but 8082

content/docs/quickstart/quickstart.sh
jetzlstorfer: that didn't work for me due to the port 8082. 
is that a bug in Keptn or do I have to fix it on my end?

christian-kreuzberger-dtx (approved):
LGTM

PR #830, opened 2021-06-01
Correct the selector of the api-service

agrimmer:
The selector of the api-service has changed from `run=api-service` to `app.kubernetes.io/name: api-service`, see https://github.com/keptn/keptn/blob/master/installer/manifests/keptn/charts/control-plane/templates/core.yaml#L52


johannes-b (approved):
lgtm

PR #833, opened 2021-06-02
keeping the announcment on keptn.sh

jetzlstorfer:
I believe it would be a good idea to keep the announcements on keptn.sh instead of linking to github


johannes-b (approved):
lgtm

PR #867, opened 2021-07-06
update Prometheus integration

jetzlstorfer:
- Update to use latest Prometheus service 
- Remove Prometheus SLi service as it is now part of the Prometheus service anyway


johannes-b (approved):
lgtm

PR #898, opened 2021-08-12
Adding missing docu for OAuth/OpenID Authentication in Keptn Bridge

johannes-b:


jetzlstorfer:
Disclaimer: I am not an expert on OAuth, however, there might be some inconsistencies in the docs.

In one example the endpoint is defined as `token` while in the other example it is defined as `token_decision`:
image



Also, the payload does not match:
image

johannes-b:
@jetzlstorfer thanks for the feedback. I have fixed the inconsistency.

jetzlstorfer (approved):
lgtm

PR #950, opened 2021-10-21
Added filtering feature to integrations page

TannerGilbert:


christian-kreuzberger-dtx:
We decided to not use the notification filter for now, as it would only display one service (slack-service), as the other ones are deprecated.

PR #999, opened 2021-12-23
Provide hint to do Helm upgrade

johannes-b:
Signed-off-by: Johannes <johannes.braeuer@dynatrace.com>


oleg-nenashev (approved):

oleg-nenashev (approved):
As mentioned in Slack by @bacherfl  and confirmed by @mowies, `keptn set config KubeContextCheck true` could be also used. Merging this one so that the issue is referenced

PR #1001, opened 2021-12-28
Disable the public demo

oleg-nenashev:
Until https://github.com/keptn/community/issues/89 is ready, we cannot serve the public demo.

Tutorials are also updated by https://github.com/keptn/tutorials/pull/216


grabnerandi (approved):
lgtm

PR #1029, opened 2022-02-03
fix: Fix auto generation of CLI docs for new Keptn releases

mowies:
### This PR
- fixes an issue where the CLI docs for a new release could not be generated when a new release of Keptn happened
- removes unneeded files for the spell checker pipeline that is not used anymore
- instead of downloading the CLI using get.keptn.sh, it's now downloaded directly from the CI run in keptn/keptn which was triggered by the tag of a new release

### Related issues:
Fixes keptn/keptn#6716


.github/workflows/auto-update.yml
```diff
@@ -21,41 +16,71 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
         with:
-          fetch-depth: 0
+          token: ${{ secrets.KEPTN_BOT_TOKEN }}
+
+      - name: Checkout
```
thisthat, line 21: why two checkouts in a row?

.github/workflows/auto-update.yml
```diff
@@ -21,41 +16,71 @@ jobs:
...
+          mv "$CLI" ./keptn
+          chmod +x ./keptn
 
       - name: Generate docs
         env:
           SOURCE_VERSION_SLUG: ${{ env.SOURCE_VERSION_SLUG }}
+          DOCS_PATH: "./content/docs/${{ env.SOURCE_VERSION_SLUG }}/reference/cli/commands"
         run: |
-          full_path="./content/docs/$SOURCE_VERSION_SLUG/reference/cli/commands"
           if [[ ! -d $full_path ]]; then
```
thisthat, line 68:
```suggestion
          if [[ ! -d $DOCS_PATH ]]; then
```

.github/workflows/auto-update.yml
mowies: one is for the keptn.github.io repo and one is for the keptn/keptn repo
see: https://github.com/keptn/keptn.github.io/pull/1029/files#diff-fc41c8132b450c9cc5a4d687b8fad2774fbddc829b368b97888186542adf3d78R24

.github/workflows/auto-update.yml
mowies: good catch!

thisthat (approved):

PR #1034, opened 2022-02-11
fix broken xref to shipyard config

StackScribe:


oleg-nenashev (approved):
Thanks! There is a ticket for link consistency check automation

PR #1042, opened 2022-02-15
Add info about doc source structure, etc

StackScribe:

I added some information about things that took me a while to unravel.


johannes-b (approved):
Looks good to me!

oleg-nenashev (approved):
The content looks good! We will need to re-integrate -patches from @AnonymousCodes911 after the merge, but right now the DCO is not passing there. @StackScribe it would be great if you could follow-up and help to get it merged

PR #1067, opened 2022-02-21
Add redirect to the GSoC website

oleg-nenashev:
Redirects keptn.sh/gsoc to https://github.com/keptn/community/blob/main/mentorship/gsoc where all the content resides at the moment

Signed-off-by: Oleg Nenashev <o.v.nenashev@gmail.com>

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -12,6 +12,6 @@ HUGO_ENABLEGITINFO = "true"
 [[redirects]]
   from = "/gsoc"
   to = "https://github.com/keptn/community/blob/main/mentorship/gsoc"
-  status = 301
+  status = 302
   force = false
   query = {path = ":path"}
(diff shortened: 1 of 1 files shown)


johannes-b:
@oleg-nenashev your changes look good to me, but what is actually the reasoning why you would like to have this redirect? From where is the user coming?

oleg-nenashev:
> @oleg-nenashev your changes look good to me, but what is actually the reasoning why you would like to have this redirect? From where is the user coming?

It is for a link on GSoC public materials and in the org profile. The redirect should be helpful if we decide to move the content over to the website

mowies (changes requested):
The redirect doesn't seem to work in the preview...

oleg-nenashev:
> The redirect doesn't seem to work in the preview...

Indeed...

PR #1075, opened 2022-03-05
feat: Added Webhook filter to integrations page

### LINKED ISSUE #1036: Add an integrations type filter to the Integrations Page
In this page https://keptn.sh/docs/integrations/ I would like to have a filter for "webhook" and "service".  This way one can more easily find integrations.

TannerGilbert:
Currently filtering is done by using the `ts_query_web`, which filters using a full text search. I just asked the Artifacthub team if it's also possible to filter by annotations like `keptn/kind` or `keptn/version`. If this is possible I'll update the code appropriately.

Preview: https://deploy-preview-1075--keptn.netlify.app/docs/integrations/

image

Fixes #1036


TannerGilbert:
Update: Filtering by annotations is currently not possible. We'll need to continue using `ts_query_web` until this feature is implemented.

PR #1078, opened 2022-03-08
thekindkeptn 0.0.8 (keptn 0.13.2)

agardnerIT:


content/docs/quickstart/_index.md
```diff
@@ -21,15 +23,16 @@ Run the Keptn Hello, World! example:
...
 ```
 
 The Keptn Bridge (UI) will be available on `http://localhost`
 
 ### Outcome
 
 The command above has:
-- Created a Kubernetes cluster in Docker using [kind](https://kind.sigs.k8s.io/)
+
+- Created a Kubernetes cluster in Docker using [k3d](https://k3d.io/)
```
oleg-nenashev, line 35: Hm?

agardnerIT: It never actually was using kind. Originally, way back when I started the project (image `0.0.1`) it used kind, but kind was too flaky so since then it has used `k3d` inside the docker image. The first thing the install_script does is a `k3d cluster create`: https://github.com/agardnerIT/thekindkeptn/blob/e870cb97b23456d32e99f07e2831673891383d02/install_script.sh#L36

`thekindkeptn` can now be seen more as marketing. As in: It is kind to you - it makes it easy to spin up Keptn :)

PR #1091, opened 2022-03-09
Delete code of conduct

iamrajiv:


oleg-nenashev (commented):
I would just delete this file. Then the metadata will be picked up from the .github repository automatically

oleg-nenashev (approved):
Thanks!

Issue #1071, opened 2022-02-28
Document using DCO in the contributing guide


oleg-nenashev:
Currently it is not trivial to use DCO, especially when editing from the web browser

Spin off of https://github.com/keptn/keptn.github.io/pull/1060 . See the comments from @mowies and @StackScribe

AnonymousCodes911:
Hey @oleg-nenashev, I would like to work on this. I would really appreciate if @StackScribe can brief me on how to fix the DCO as I tried the last time and I wasn't able to get through. 
Thankyou!

PR #1095, opened 2022-03-10
docs: Add doc for DCO

iamrajiv:
Fixes: #1071

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -15,57 +15,58 @@ We were all new to this once and are happy to help you!
 
 ## Guidelines for contributing
 
-* Always fork the repository then clone that fork to your local system rather than clone master directly.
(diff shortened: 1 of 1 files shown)


oleg-nenashev (changes requested):
Thanks for submitting it! I believe we need to put some attribution to the sources of the text. 

The text copies text from other projects, e.g.  https://github.com/opensearch-project/common-utils/blob/main/CONTRIBUTING.md or https://code.asam.net/simulation/wiki/-/wikis/docs/project_guidelines/ASAM-DCO?version_id=c510bffb1195dc04deb9db9451112669073f0ba5 or https://thesofproject.github.io/latest/contribute/contribute_guidelines.html

iamrajiv:
@oleg-nenashev I have added the attributions.

oleg-nenashev (approved):
Thanks! Let's merge it here.
Later we might want to move it to a project global contributing guide

PR #1101, opened 2022-03-11
fix: Keptn branding assets URL

iamrajiv:


iamrajiv:
@oleg-nenashev Can you check this too?

oleg-nenashev (approved):
Right, thanks

PR #1102, opened 2022-03-12
Created 'Files' subsection in reference section

Ciggzy1312:
Signed-off-by: Deepayan <deepayanmukherjee1312@gmail.com>

Moved sli slo sli-provider to files subsection in reference section


StackScribe (commented):
I think this looks good and we need to do this step first so we can start adding other file reference pages to the set.

One caveat: we may need to modify the metrics field in the index.md file for each of these.  They have values like 10 and 15 and s falls pretty late in the alphabet.  I almost wonder if we want to start a much larger range -- maybe give these 7xx values so we have plenty of room to maneuver.  wdyt?

johannes-b:
Hi @Ciggzy1312, 

thanks for the PR. Please take into account that there is already a folder for the upcoming `0.14.x` release. Hence I would like to ask you to perform the same changes in `0.14.x` (https://github.com/keptn/keptn.github.io/tree/master/content/docs/0.14.x) as well to not lose them in the future. Thanks!

FYI: @StackScribe

oleg-nenashev:
Time to merge. I do not see external links

StackScribe:
I agree that it's time to merge this.  I'm thinking that we should merge the change only into 0.14.x, no 0.13.x.  Then we can build up this section and roll it out when 0.14.x is released.  We need to make sure we catch xrefs to these files and so forth.

content/docs/0.14.x/reference/files/_index.md
```diff
@@ -0,0 +1,4 @@
+---
+title: Files
+description: Definition of SLIs/SLOs of quality gate and how to customize them.
```
StackScribe, line 3:
```suggestion
description: Reference pages for Keptn files.
```
The sli/slo pages are only the beginning here.  This section will soon include pages for many other files.

StackScribe (changes requested):

StackScribe:
I can't get into the slo/index.md and sli/index.md files to see what the weight value is but, if it is unchanged from the weight used when these files were in the quality_gates tree, I think the values are way too low.  This is the beginning of a new section that is going to end up with a lot of files that need to be listed alphabetically and "s" falls fairly late in the alphabet...  Can someone check/fix this?

oleg-nenashev (approved):

PR #1103, opened 2022-03-12
Moved keptn brigde to "Bridge' section on same level as references

Ciggzy1312:
Signed-off-by: Deepayan <deepayanmukherjee1312@gmail.com>

-> Moved Keptn bridge under "Bridge" subsection
-> Bridge subsection is kept on same level as Quality Gates and References as per docs proposal

-> The description for the Bridge section is yet to be decided
-> Implementation of "Bridge" entry to Glossary and Concepts article are yet to be discussed


StackScribe (commented):
Good idea to just get the Bridge files out of the Reference section for now.  But I think we want to put them in a section on the same level as Quality Gates, so within the release-specific section.  That will retain the general shape of the current doc set until we revamp the whole thing.

I think it is okay to pull the Reference section out of the release-specific section, so it will be at the same level as Concepts.  That is just a little tweak in the structure  ;-). Other opinions?

content/docs/0.13.x/Bridge/_index.md
```diff
@@ -0,0 +1,3 @@
+---
+title: Bridge
```
oleg-nenashev, line 2:
```suggestion
title: Bridge (Web UI)
```

oleg-nenashev (approved):
No objections from me. Maybe it makes sense to add a second term for Bridge which is more friendly to newcomer users

StackScribe:
I just found https://keptn.sh/docs/concepts/architecture/#keptn-bridge -- needs some work sometime but it does link to the Bridge pages that are currently in the Reference section.  Those links should be replaced with links to this new section in this PR.

johannes-b:
Hi @Ciggzy1312, 

thanks for the PR. Please take into account that there is already a folder for the upcoming `0.14.x` release. Hence I would like to ask you to perform the same changes in `0.14.x` (https://github.com/keptn/keptn.github.io/tree/master/content/docs/0.14.x) as well to not lose them in the future. Thanks!

FYI: @oleg-nenashev & @StackScribe

PR #1110, opened 2022-03-16
Fix:fix a typo and refactored MACOS to macOs

Biki-das:
MacOS should be macOS

`macOS` is a proprietary graphical operating system developed and marketed by Apple Inc. since 2001. It is the primary operating system for Apple's Mac computers. [Wikipedia](https://en.wikipedia.org/wiki/MacOS)


oleg-nenashev (approved):
Thanks!

PR #1122, opened 2022-03-17
add intro to install page

StackScribe:

Added a bit of contextual, intro text.


oleg-nenashev (changes requested):
0.14.x branch would need to be updated before merging

oleg-nenashev (approved):

Issue #1166, opened 2022-04-05
0.14.x contents in left frame is out of order


StackScribe:
I see the following problems:
* Bridge section is at top level of the release directory and has no weight (it is playing out under Troubleshooting, oddly enough) in the left frame
* 0.13.x and 0.14.x have the same weight

Need to fix these issues and see if that makes the release sections appear in the proper order -- 0.14.x, 0.13.x, 0.12.x -- currently, 0.12.x is way out of place.

@StackScribe is working on this

PR #1167, opened 2022-04-05
Fix 0.14.x structure issues

StackScribe:

https://github.com/keptn/keptn.github.io/issues/1166

* changed weight of 0.14.x to 983 (one less than weight of 0.13.x)
* added weight to Bridge section (although that section is superfluous and was subsequently deleted but I just wanted to see how much of the mess would be fixed with that change)
* moved the Bridge/bridge tree to be a top-level directory under 0.14.x, located between "Continuous Delivery" and "Quality Gates"
* Note that this seems to have recovered the missing sequence screen shot that Rob Jahn reported although I did not make any changes to fix that  
* in reference/cli, renamed _index.md to index.md.  This fixed the problem of the  CLI section showing as > CLI under API in the left frame TOC

This PR leaves some issues in the Bridge section.  Previously, this material was in the "Reference" section, where it did not belong, so this consolidates it in its own section:
* 5 of the 7 subsections have weight 40 so may not be appearing in the proper order but I don't know what the proper order is so I  left it alone.  At least all the files are together in one section that is entitled "Keptn Bridge" so someone can fix the ordering in a separate PR.
* I have not scrubbed through the other docs to check for xrefs that need to be fixed.  That can be done in a separate PR.


oleg-nenashev (approved):
The preview looks good, thanks!

PR #1177, opened 2022-04-08
[WIP] Create shipyard.yaml reference file

StackScribe:
TODO:
* Add info about properties
* Find other information that is lurking out there
* Fix xrefs to and from this document
* Figure out where the See also info not coded in this file originates
* AAACK!  I have previous branches pulled into this PR!  Not quite sure how I did that but it must be resolved.  I expect this PR to die and be replaced by another one that has only the shipyard material!

@johannes-b If you have a minute, could you just glance at this -- it's not yet ready for a detailed review.  But I'm thinking that this is getting to be way too big for a single document and it does not yet contain all the information.   What would you think about a reference "section" for a single file?  Perhaps the top-level file would contain the very basic info about a shipyard file -- that it has stages, sequences, tasks, properties, etc -- with each term linking to a sub-page that just discusses that piece?  Or maybe you have other ideas.  I think it would be good to consolidate all the information about the shipyard file structure into a single place but maybe that doesn't need to be a single massive file?


StackScribe:
This PR is replaced by https://github.com/keptn/keptn.github.io/pull/1178 .

PR #1178, opened 2022-04-08
[WIP] Create reference page for shipyard.yaml

StackScribe:

This commit is being developed in 0.15.x but will be moved to 0.16.x when possible.  It is too big and too disruptive to merge in a published version of the docs.


StackScribe:
Not really WIP.  It could be mergeable -- I'd like other opinions here.  My initial thought was to just take the existing "Shipyard" file and call it a reference page then clean it up later.  But I'm thinking that this may be unwieldy -- could subdivide this or maybe put much of this back in the Quality Gates section as a guide and do a real reference page here.

PR #1226, opened 2022-05-16
Update of Kubernetes support matrix

johannes-b:
Signed-off-by: Johannes <johannes.braeuer@dynatrace.com>

* Update of K8s support matrix
* Removed suspect footnote


thisthat (approved):

StackScribe (commented):
* For Remarks(1), do we really want to call out sockshop at this point?  Can we give any sizing recommendations for someone trying to set up a production Keptn installation?  
* Note 3 -- do we want to end the sentence after "might work".  Then add "We have had some success running Keptn for study and demonstrations on k3s and k3d (which could like to the instructions) but do not guarantee that it works in all cases."  And then the "Please refer to..." sentence as it is.
* Under "Cluster size", how about making the link say "See the `Kubernetes Resource Data" information towards the bottom of the section for each release in the https://github.com/keptn/keptn/releases Release Notes."  It is a little bit hidden -- now I do see it but, with no explanation, I expect the link to take me right to the information.

I guess we don't yet have the experience to do much more about how to size a cluster -- like how to sum up the information here to figure out a reasonable size, right?  Is there any more information we can give?  Like maybe spell out the storage needs for Persistent Volumes -- it seems like we're seeing a fair number of people who can't get Keptn up and running because their cluster does not have adequate space for the persistent volumes that are required  for the services they are using.  Ugh!  This is tough stuff!

PR #1234, opened 2022-05-19
[WIP] shipyard.yaml reference file

StackScribe:

This content was developed in 0.15.x.  The https://github.com/keptn/keptn.github.io/pull/1238 PR moves this content to 0.16.x and replaces this PR.

This PR does the following:
* Creates a reference page for the shipyard.yaml file to summarize the synopsis of the file.  We will certainly refine this page more and add xrefs to this file from other pages as we move forward but, if reviewers agree, this is the general shape of the page.
* Deletes old material from the manage/shipyard file that is now in this file.
* Creates a separate manage/triggers file.  Note that this is unchanged from what was previously in the manage/shipyard file so does not require review for this exercise.

Issues for review:
* Should this page be called "shipyard.yaml" or just "shipyard"?
* Are the elements in this file "fields" or "parameters" or something else?
* What should be included in the "Synopsis" section?  I tried to incorporate some tasks into the Synopsis but not sure if this works and exactly how to arrange them.
* How should we handle "optional" bits in the Synopsis?  I started to surround them with square brackets (e.g. [triggeredOn...] following man page conventions but I fear that may just muddle the issue, especially for a yaml file where indentation is so significant.  Opinions?
* Should we include the mini-synopses in each task section or should we get all of that in the main "Synopsis" section?  If we keep them in the individual task section, is having the "Synopsis:" lead-in helpful or just noise?
* What should be on this page that is not currently included?  I am favoring keeping this terse with xrefs to other pages in the doc set for more details and examples but perhaps there is other information that should be included here.

This PR replaces https://github.com/keptn/keptn.github.io/pull/1178.


StackScribe:
This content has been moved to https://github.com/keptn/keptn.github.io/pull/1238, which implements the changes in the 0.16.x tree.  I am closing this PR.

PR #1238, opened 2022-05-24
add shipyard ref, manage triggers, modify manage shipyard

StackScribe:

This PR does the following:
* Creates a reference page for the shipyard.yaml file to summarize the synopsis of the file.  We will certainly refine this page more and add xrefs to this file from other pages as we move forward but, if reviewers agree, this is the general shape of the page.
* Deletes old material from the manage/shipyard file that is now in this file.
* Creates a separate manage/triggers file.  Note that this is unchanged from what was previously in the manage/shipyard file so does not require review for this exercise.

Issues for review:
* Should this page be called "shipyard.yaml" or just "shipyard"?
* Are the elements in this file "fields" or "parameters" or something else?
* What should be included in the "Synopsis" section?  I tried to incorporate some tasks into the Synopsis but not sure if this works and exactly how to arrange them.
* How should we handle "optional" bits in the Synopsis?  I started to surround them with square brackets (e.g. [triggeredOn...] following man page conventions but I fear that may just muddle the issue, especially for a yaml file where indentation is so significant.  Opinions?  **Update**: @agardnerIT suggested we use comments to show what is optional so I tried that for a couple lines.  I'm not sure I like the results.  Perhaps we should not worry about showing what is optional in the Synopsis?  Almost everything in shipyard is optional, depending on what one is doing with the project.  Perhaps we should just explain what is optional in the prose description of each line.
* Should we include the mini-synopses in each task section or should we get all of that in the main "Synopsis" section?  If we keep them in the individual task section, is having the "Synopsis:" lead-in helpful or just noise?
* What should be on this page that is not currently included?  I am favoring keeping this terse with xrefs to other pages in the doc set for more details and examples but perhaps there is other information that should be included here.

This PR adds content to the unreleased 0.16.x tree and replaces https://github.com/keptn/keptn.github.io/pull/1234/

### DIFF
--- a/content/docs/0.16.x/manage/shipyard/index.md
+++ b/content/docs/0.16.x/manage/shipyard/index.md
@@ -2,7 +2,7 @@
 title: Working with shipyard files
 description: Information about shipyard, sequences and tasks to define processes and workflows.
 weight: 25
-keywords: [0.15.x-manage]
+keywords: [0.16.x-manage]
 aliases:
 ---
 
@@ -51,8 +51,7 @@ This means that all services in a project share the same shipyard definition.
             - name: "deployment"
             - name: "release"
 
-        - name: "production"
-          sequences:
+
           - name: "delivery"
             triggeredOn:
               - event: "hardening.delivery.finished"
--- a/content/docs/0.16.x/manage/shipyard/index.md
+++ b/content/docs/0.16.x/manage/shipyard/index.md
@@ -1,290 +1,103 @@
 ---
-title: Shipyard
+title: Working with shipyard files
 description: Information about shipyard, sequences and tasks to define processes and workflows.
 weight: 25
-keywords: [0.16.x-manage]
+keywords: [0.15.x-manage]
 aliases:
 ---
 
-## Declare Shipyard (before creating a project)
-
-* A shipyard is defined at the level of a project. This means that all services in a project share the same shipyard definition.
+A [shipyard.yaml](../../reference/files/shipyard) file is defined at the level of a project.
+This means that all services in a project share the same shipyard definition.
 
 * A shipyard defines the stages each deployment has to go through until it is released in the final stage, e.g., the production stage.
 
-* A shipyard can consist of any number of stages; but at least one. A stage must have at least the name property.
+* A shipyard can consist of any number of stages; but at least one. Each stage must have at least the name property.
 
-* A stage can consist of any number of sequences; but at least one.
-
-
-### Definition of Stage
-
-A stage is declared by its name. This name is used for the branch in the Git repository and Kubernetes namespace to which services at this stage will be deployed.
+* A stage can contain any number of sequences but must have at least one.
 
 **Example of a shipyard with three stages:**
 
-```yaml
-apiVersion: spec.keptn.sh/0.2.3
-kind: "Shipyard"
-metadata:
-  name: "shipyard-sockshop"
-spec:
-  stages:
-    - name: "dev"
-    - name: "hardening"
-    - name: "production"
-```
-
-### Definition of Sequence in a Stage
-
-After defining stages, sequences can be added to a stage. A sequence is an ordered list of tasks that are triggered sequentially. A `sequence` consists of the following properties:
-
-* `name`: A unique name of the sequence
-* `triggeredOn` (optional): An array of events that trigger the sequence.
-* `tasks`: An array of tasks executed by the sequence in the declared order.
+    apiVersion: spec.keptn.sh/0.2.3
+    kind: "Shipyard"
+    metadata:
+      name: "shipyard-sockshop"
+    spec:
+      stages:
+        - name: "dev"
+        - name: "hardening"
+        - name: "production"
 
 **Example:** Extended shipyard with a delivery sequence in all three stage:
 
-```yaml
-apiVersion: spec.keptn.sh/0.2.3
-kind: "Shipyard"
-metadata:
-  name: "shipyard-sockshop"
-spec:
-  stages:
-    - name: "dev"
-      sequences:
-      - name: "delivery"
-        tasks: 
-        - name: "deployment"
-        - name: "release"
-
-    - name: "hardening"
-      sequences:
-      - name: "delivery"
-        triggeredOn:
-          - event: "dev.delivery.finished"
-        tasks: 
-        - name: "deployment"
-        - name: "release"
-
-    - name: "production"
-      sequences:
-      - name: "delivery"
-        triggeredOn:
-          - event: "hardening.delivery.finished"
-        tasks: 
-        - name: "deployment"
-        - name: "release"
-```
-
-### Reserved Keptn Tasks
-
-Keptn supports a set of opinionated tasks for declaring a delivery or remediation sequence:
-
-* action
-* approval
-* deployment
-* evaluation
-* get-action
-* release
-* rollback
-* test
-
-#### Action
-
-The action task indicates that a remediation action should be executed by an action provider. It is used within a [remediation workflow](../../automated_operations/remediation).
-
-**Usage:**
-
-```
-- name: action
-```
-
-#### Approval
-
-The approval task intercepts the task sequence and waits for a user to approve or decline the open approval. This task can be added, for example, before deploying an artifact into the next stage. The approval strategy can be defined based on the evaluation result `pass` and `warning`. Keptn supports the following approval strategies for the evaluation results `pass` and `warning`:
-
-* `automatic`: The artifact is deployed automatically.
-* `manual`: The user is asked for approval before triggering the deployment.
-
-This allows combinations as follows:
-
-|                          | Evaluation result: pass           | Evaluation result: warning                 | Behavior  |
-|--------------------------|-----------------------------------|--------------------------------------------|-----------|
-| **Skip approval task:** | pass:automatic | warning:automatic | Regardless of the evaluation result, the approval task is skipped |
-| **Depending on evaluation result:**   | pass:automatic | warning:manual    | If the evaluation result is a **warning**, an approval is required |
-| **Depending on evaluation result:**   | pass:manual    | warning:automatic | If the evaluation result is a **pass**, an approval is required |
-| **Mandatory approval task:**          | pass:manual    | warning:manual    | Regardless of the evaluation result, an approval is required |
-
-By default, an `automatic` approval strategy is used for evaluation result `pass` and `warning`.
-
-**Usage:**
-
-```
-- name: approval
-  properties: 
-    pass: automatic
-    warning: manual
-```
-
-**Note:** If the approval is the first task of a sequence, the approval service is not provided with the result from a previous task. In this case, it always falls back to the `manual` approval strategy.
-
-<details><summary>**Example:** Extended shipyard with a mandatory approval task in production</summary>
-
-<p>
-
-```yaml
-apiVersion: spec.keptn.sh/0.2.3
-kind: "Shipyard"
-metadata:
-  name: "shipyard-sockshop"
-spec:
-  stages:
-    - name: "production"
-      sequences:
-        - name: "delivery"
-          tasks:
+    apiVersion: spec.keptn.sh/0.2.3
+    kind: "Shipyard"
+    metadata:
+      name: "shipyard-sockshop"
+    spec:
+      stages:
+        - name: "dev"
+          sequences:
+          - name: "delivery"
+            tasks: 
             - name: "deployment"
-            - name: "approval"
-              properties:
-                pass: "manual"
-                warning: "manual"
             - name: "release"
-```
 
-</p>
-</details>
+        - name: "hardening"
+          sequences:
+          - name: "delivery"
+            triggeredOn:
+              - event: "dev.delivery.finished"
+            tasks: 
+            - name: "deployment"
+            - name: "release"
 
-#### Deployment
-
-Defines the deployment strategy (see [Continuous Delivery](../../continuous_delivery/)) used to deploy a new version of a service. Keptn supports deployment strategies of type:
-
-* `direct`: Deploys a new version of a service by replacing the old version of the service.
-* `blue_green_service`: Deploys a new version of a service next to the old one. After a successful validation of this new version, it replaces the old one and is marked as stable (i.e., it becomes the `primary`-version).
-* `user_managed` (experimental): Deploys a new version of a service by fetching the current helm-chart from the git repo, and just updating certain values.
-
-**Usage:**
-
-```
-- name: deployment
-  properties: 
-    deploymentstrategy: blue_green_service
-```
-
-```
-- name: deployment
-  properties: 
-    deploymentstrategy: direct
-```
-
-```
-- name: deployment
-  properties: 
-    deploymentstrategy: user_managed
-```
-
-#### Evaluation
-
-Defines the quality evaluation that is executed to verify the quality of a deployment based on its SLOs/SLIs.
-
-**Usage:**
-```
-- name: evaluation
-```
-
-#### Get-Action
-
-The get-action task is used to extract the desired remediation action from a remediation.yaml within a [remediation workflow](../../automated_operations/remediation).
-
-**Usage:**
-
-```
-- name: get-action
-```
-
-#### Release
-
-Defines the releasing task that is executed after a successful deployment happens. This means that production traffic is shifted towards the new deployment in this task.
-
-**Usage:**
-```
-- name: release
-```
-
-#### Action
-
-Defines the execution of a remediation action retrieved by `get-action`.
-
-**Usage:**
-```
-- name: action
-```
-
-#### Rollback
-
-Defines the rollback task that is executed when a rollback is triggered.
-
-**Usage:**
-```
-- name: rollback
-```
-
-#### Test
-
-Defines the test strategy used to validate a deployment. Keptn supports tests of type:
-
-* `functional`: Test a deployment based on functional tests.
-* `performance`: Test a deployment based on performance/load tests.
-
-**Usage:**
-```
-- name: test
-  properties: 
-    teststrategy: functional
-```
+        - name: "production"
+          sequences:
+          - name: "delivery"
+            triggeredOn:
+              - event: "hardening.delivery.finished"
+            tasks: 
+            - name: "deployment"
+            - name: "release"
 
 <details><summary>**Example:** Extended shipyard with functional tests in dev and performance tests in hardening
 </summary>
 
 <p>
 
-```yaml
-apiVersion: spec.keptn.sh/0.2.3
-kind: "Shipyard"
-metadata:
-  name: "shipyard-sockshop"
-spec:
-  stages:
-    - name: "dev"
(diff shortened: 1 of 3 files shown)


content/docs/0.16.x/manage/shipyard/index.md
```diff
@@ -1,290 +1,103 @@
 ---
-title: Shipyard
+title: Working with shipyard files
 description: Information about shipyard, sequences and tasks to define processes and workflows.
 weight: 25
-keywords: [0.16.x-manage]
+keywords: [0.15.x-manage]
```
oleg-nenashev, line 5: It is `0.16.x`

oleg-nenashev (commented):
> Should this page be called "shipyard.yaml" or just "shipyard"?

I would prefer `shipyard`, but other files seems to be YAML



> How should we handle "optional" bits in the Synopsis? I started to surround them with square brackets (e.g. [triggeredOn...] following man page conventions but I fear that may just muddle the issue, especially for a yaml file where indentation is so significant. Opinions?

Works for me

> Should we include the mini-synopses in each task section or should we get all of that in the main "Synopsis" section? If we keep them in the individual task section, is having the "Synopsis:" lead-in helpful or just noise?

I would keep everything in Synopsis for now to avoid duplication. We do not have macros

> What should be on this page that is not currently included? I am favoring keeping this terse with xrefs to other pages in the doc set for more details and examples but perhaps there is other information that should be included here.

content/docs/0.16.x/manage/shipyard/index.md
```diff
@@ -1,290 +1,103 @@
...
-* A shipyard is defined at the level of a project. This means that all services in a project share the same shipyard definition.
+A [shipyard.yaml](../../reference/files/shipyard) file is defined at the level of a project.
+This means that all services in a project share the same shipyard definition.
 
 * A shipyard defines the stages each deployment has to go through until it is released in the final stage, e.g., the production stage.
 
-* A shipyard can consist of any number of stages; but at least one. A stage must have at least the name property.
-
-* A stage can consist of any number of sequences; but at least one.
-
```
agardnerIT, line 18: * A sequence can consist of any number of tasks; but at least one.

content/docs/0.16.x/manage/shipyard/index.md
```diff
@@ -1,290 +1,103 @@
...
-```
-
-### Definition of Sequence in a Stage
-
-After defining stages, sequences can be added to a stage. A sequence is an ordered list of tasks that are triggered sequentially. A `sequence` consists of the following properties:
-
-* `name`: A unique name of the sequence
-* `triggeredOn` (optional): An array of events that trigger the sequence.
-* `tasks`: An array of tasks executed by the sequence in the declared order.
+    apiVersion: spec.keptn.sh/0.2.3
```
agardnerIT, line 20: I know what you're getting at but this is misleading IMO. Stages without sequences and tasks makes no sense and won't work (by that I mean the user couldn't use the shipyard below to actually "achieve" anything)

content/docs/0.16.x/manage/shipyard/index.md
```diff
@@ -1,290 +1,103 @@
...
+    kind: "Shipyard"
+    metadata:
+      name: "shipyard-sockshop"
+    spec:
+      stages:
+        - name: "dev"
+        - name: "hardening"
+        - name: "production"
 
 **Example:** Extended shipyard with a delivery sequence in all three stage:
```
agardnerIT, line 30: 1. "Extended" feels like an "optional extra" which, given the comment above, it isn't.
2. tasks: block cannot be empty. A sequence must include at least one task. The task name can technically be the same as the sequence name but it leads to confusion with users who think they trigger tasks (they don't; external tools and humans only ever trigger at the sequence level. Tools are triggered by tasks).

content/docs/0.16.x/manage/shipyard/index.md
```diff
@@ -1,290 +1,103 @@
...
-      sequences:
-      - name: "delivery"
-        triggeredOn:
-          - event: "hardening.delivery.finished"
-        tasks: 
-        - name: "deployment"
-        - name: "release"
-```
-
-### Reserved Keptn Tasks
```
agardnerIT, line 81: Excellent! The first time I've seen this anywhere!

content/docs/0.16.x/manage/shipyard/index.md
```diff
@@ -1,290 +1,103 @@
...
-Keptn supports a set of opinionated tasks for declaring a delivery or remediation sequence:
-
-* action
-* approval
-* deployment
-* evaluation
-* get-action
-* release
-* rollback
-* test
```
agardnerIT, line 92: Hmmm... Technically speaking the `test` and `release` tasks are "reserved" because we deploy OOTB with the `helm-service` and `jmeter-service` because historically Keptn was opinionated on what `test` and `release` meant (ie. you WILL test with jmeter and you WILL release with helm). Fastforward to today, keptn is much less opinionated and let's users fully bring their own tooling.

Another problem with the OOTB helm-service is it's presupposition that you will use Istio (I'll discuss this further later) and as @christian-kreuzberger-dtx's POC shows, it is entirely possibly to run vanilla helm (ie. without assuming Istio is present) using the Job Executor Service).

https://github.com/keptn/keptn/issues/7845 shows that we're moving them to the `keptn-contrib` repo.

content/docs/0.16.x/manage/shipyard/index.md
```diff
@@ -1,290 +1,103 @@
...
-                warning: "manual"
             - name: "release"
-```
-
-</p>
-</details>
-
-#### Deployment
-
-Defines the deployment strategy (see [Continuous Delivery](../../continuous_delivery/)) used to deploy a new version of a service. Keptn supports deployment strategies of type:
```
agardnerIT, line 161: See notes above re: `jmeter-service` and `helm-service`. Drawing the difference between Keptn control plane and the tooling services. The control plane technically doesn't support any deployment strategies. This is old wording (and still technically true because Keptn still ships with the jmeter-service and `helm-service` in the core but (as above) this will soon no longer be the case.

All that said, actually it is the `helm-service` with it's assumption that Istio is installed on the customer cluster that really "supports" these deployment options.

content/docs/0.16.x/manage/shipyard/index.md
```diff
@@ -1,290 +1,103 @@
...
-Defines the rollback task that is executed when a rollback is triggered.
-
-**Usage:**
-```
-- name: rollback
-```
-
-#### Test
-
-Defines the test strategy used to validate a deployment. Keptn supports tests of type:
```
agardnerIT, line 235: It is only the `jmeter-service` that supports the `functional` and `performance` keywords. Indeed task properties are:
a) Entirely optional
b) Send for, processed by and only relevant for the tool processing the task event.

In other words, if you asked k6 to respond to a `test` event with a property set to `performance` it would be entirely ignored.

Perhaps move "task properties" to it's own page and entirely remove them from this page (ie. leave all shipyard files sans properties block)?

content/docs/0.16.x/manage/shipyard/index.md
```diff
@@ -456,211 +269,3 @@ spec:
 
 **Result:** After extending the shipyard as shown above, remediations should be executed when a problem event is retrieved (see [remediation workflow](../../automated_operations/remediation)).
 
-### Define a trigger for a sequence 
-
-An advanced and powerful feature of the shipyard is that you can define *triggers* to kick-off a sequence. Therefore, a sequence offers the `triggeredOn` property where a list of events can be specified. The event types you can list there are events that refer to the status of a sequence execution. Their name follows the pattern:
```
agardnerIT, line 461: By default Keptn sequences are standalone. They run and they finish. An advanced and powerful feature of the shipyard is that you can *link* sequences together to form chains of sequences. This can be achieved using the `triggeredOn` property....

content/docs/0.16.x/manage/triggers/index.md
```diff
@@ -0,0 +1,231 @@
+---
+title: Triggers
+description: Using triggers to kick off a sequence
+weight: 50
+keywords: [0.15.x-manage]
+aliases:
+---
+
+You can define *triggers* to kick off a sequence in response to specific events.
```
agardnerIT, line 9: Other OSS tools use the word "trigger" to denote starting from an external source.

You can begin Keptn sequences via the bridge UI, API or CLI. This page specifically talks about an internal sequence-to-sequence triggering mechanism _inside_ keptn.

content/docs/0.16.x/manage/shipyard/index.md
```diff
@@ -1,290 +1,103 @@
...
-| **Depending on evaluation result:**   | pass:manual    | warning:automatic | If the evaluation result is a **pass**, an approval is required |
-| **Mandatory approval task:**          | pass:manual    | warning:manual    | Regardless of the evaluation result, an approval is required |
-
-By default, an `automatic` approval strategy is used for evaluation result `pass` and `warning`.
-
-**Usage:**
-
-```
-- name: approval
-  properties:
```
agardnerIT, line 126: This is 100% correct but FYI: Not all tasks (tools actioning those tasks) require properties so `properties` is optional. In fact, most tools do not use properties. Even those that do typically will have their properties passed in at runtime via JSON data so as to not hardcode it into the shipyard file.

For example:

```
tasks
  - name: evaluation
    properties:
      timeframe: 5m
```
Denotes that we want to run a quality gate evaluation and look at the previous 5 minutes.

This duration is hardcoded into the shipyard file. When keptn generates a cloudevent for the `evaluation.triggered` event, it comes out like this:

```
{
  "id": "1234...",
  "type": "sh.keptn.event.evaluation.triggered",
  # other fields removed for brevity
  "data": {
     "evaluation": {    # notice this is the task name
        "timeframe": "5m"
     }
}
}
```

So you could just as easily pass this in when you trigger the sequence via curl:

```
# JSON payload which triggers the sequence from outside keptn
# Notice the evaluation timeframe is already given by the user
# Meaning the properties block in the shipyard is unneccessary
# Benefit here is that the timeframe is now dynamic
{
  "type": "sh.keptn.event.SomeStage.MySequence.triggered",
  # Other fields removed for brevity
  "data": {
    "evaluation": {
      "timeframe": "5m"
    }
  }
}
```

content/docs/0.16.x/manage/shipyard/index.md
```diff
@@ -1,290 +1,103 @@
 ---
-title: Shipyard
+title: Working with shipyard files
 description: Information about shipyard, sequences and tasks to define processes and workflows.
 weight: 25
-keywords: [0.16.x-manage]
+keywords: [0.15.x-manage]
 aliases:
 ---
```
agardnerIT, line 7: General comments for the page (not necessarily at this line):

1) It is standard and a best practice NOT to have tool names inside a shipyard file. Sequences and tasks should be tool agnostic. A shipyard describes **what** you want to achieve, not **how**.
2) External tools (and humans) trigger sequences
3) Tooling respond to task events
4) task properties are optional. `properties` is a mechanism by which you can include properties that the actioning tool requires.

content/docs/0.16.x/reference/files/shipyard/index.md
```diff
@@ -0,0 +1,303 @@
+---
+title: shipyard.yaml
+description: Control orchestation for a Keptn project
+weight: 715
+---
+
+The *shipyard.yaml* file defines the activiies to be performed for a Keptn project
```
agardnerIT, line 7: slight typo: `activiies` > `activities`

agardnerIT, line 7: slight typo: `activiies` > `activities`

content/docs/0.16.x/reference/files/shipyard/index.md
```diff
@@ -0,0 +1,303 @@
...
+and the order in which those activities are executed.
+
+Each project must have one, and only one, *shipyard.yaml* file
+which is passed [**keptn create project**](../../cli/commands/keptn_create_project/) command.
```
agardnerIT, line 11: `passed to the`

agardnerIT, line 11: `passed to the`

content/docs/0.16.x/reference/files/shipyard/index.md
```diff
@@ -0,0 +1,303 @@
...
+      name: "shipyard-<project-name>"
+    spec:
+      stages:
+        - name: "<stagename-1>"
+          sequences:
+          - name: "delivery"
+            tasks:
+            - name: "deployment"
+              properties:
+                deploymentstrategy: "direct" | "blue_greem_service" | "user_managed"
```
agardnerIT, line 40: typo: `blue_green_service` but see notes elsewhere that this `properties` block may not be necessary

agardnerIT, line 40: typo: `blue_green_service` but see notes elsewhere that this `properties` block may not be necessary

content/docs/0.16.x/manage/shipyard/index.md
StackScribe: Ha!  I actually stole this from https://keptn.sh/docs/0.15.x/manage/shipyard/#reserved-keptn-tasks -- but now we'll know where it is ;-)

content/docs/0.16.x/manage/shipyard/index.md
StackScribe: Interesting...  Perhaps it would be good to note that these are specifically for helm-service and jmeter-service.  In some future release, we will describe the tooling that is required for these, then use "Difference between versions" to say something like, "In Releases 0.17.x and earlier, these tasks were truly available without the extra tooling -- worded better, of course, and probably more details.

content/docs/0.16.x/manage/shipyard/index.md
StackScribe: Okay, so I do need to do some rewriting here.  And I guess I need to start reflecting the Job Executor Service here now.   Let me do some research.

content/docs/0.16.x/manage/shipyard/index.md
StackScribe: Fixed

content/docs/0.16.x/manage/shipyard/index.md
StackScribe: This is unmodified content from the existing docs -- the file is here because I stripped material out of it.  We're going to be mix-and-matching and rewriting all of this.

content/docs/0.16.x/manage/shipyard/index.md
StackScribe: Yeah, this all needs a radical rewrite.

content/docs/0.16.x/manage/shipyard/index.md
StackScribe: Note that this was deleted from this file -- it's in the reference page

content/docs/0.16.x/manage/shipyard/index.md
StackScribe: I modified the reference page slightly to reflect that these two tasks are associated with specific services.

content/docs/0.16.x/manage/shipyard/index.md
StackScribe: Updated this material in the reference file

content/docs/0.16.x/manage/shipyard/index.md
StackScribe: I think I handled this in the reference page, for now, and we'll have to figure out the rest.

We need a sample shipyard file that has a test task that uses something other than jmeter.  I am assuming that it cannot be named "test" because that term is reserved for jmeter, right?  I pulled out the old "for historical reasons and backwards compatibility" trope to explain why these tasks are special ;-)  And then we need to create a page in the "Creating shipyard files" section about using test tools other than jmeter -- I'm not sure whether this all needs to be defined in the service for that tool or if I have some flexibility about how I incorporate it into my shipyard -- I suspect the latter.

content/docs/0.16.x/manage/shipyard/index.md
StackScribe: Reworded in reference page

StackScribe: I also modified the "Triggers" page to begin with a list of  ways a sequence can be triggered, then put the content we have into a section entitled "Using triggeredOn in a sequence".  We need to go back and add content for the other methods and probably have other work to do but this is a start...

content/docs/0.16.x/manage/triggers/index.md
StackScribe: See previous comments.  This is fixed, sort of, for now.

content/docs/0.16.x/manage/shipyard/index.md
StackScribe: The ref file does say that properties are optional, and I added a note about the JSON

content/docs/0.16.x/reference/files/shipyard/index.md
StackScribe: Fixed

StackScribe: Fixed

content/docs/0.16.x/reference/files/shipyard/index.md
StackScribe: fixed

StackScribe: Fixed

content/docs/0.16.x/reference/files/shipyard/index.md
StackScribe: It is necessary with helm-service, right?  We'll keep working on how to represent this properly

StackScribe: See comment above

content/docs/0.16.x/manage/shipyard/index.md
StackScribe: I think this is incorporated

StackScribe:
We discussed the shipyard reference page and reference pages in general in the 2 June 2022 Community Developer Meeting and agreed to the following:
* This section should be called "Configurations" rather than "Files"
* Pages should be named without the .yaml suffix, but a sentence such as "The shipyard configuration is defined in a shipyard.yaml file."
* "Fields" is the proper header for that section.  This is the term Kubernetes uses for similar reference pages.
* The "Synopsis" section should not show anything as optional.  Depending on ones definition, almost everything in the file is optional.  The description under Field will clarify what is optional.
* All Synopsis info should be in the "Synopsis" section rather than embedding some synopsis info in the "Tasks" section.  Note that this PR leaves a bit of synopsis in the task sections because not all tasks are yet reflected in the main "Synopsis" section but this will be fixed.

These decisions conform to @oleg-nenashev 's comments.

Completing these reference pages will be an iterative process that takes some time and, as we work on them, we may decide to alter our opinions on these issues but this is a good place to start the process.

thisthat (approved):

PR #1241, opened 2022-05-30
Fix xrefs for 0.15.x

StackScribe:

Fixes some xrefs from "Concepts" articles.  Mostly, xrefs that had 0.14.x hard-coded but a couple that should point to new reference pages and the Bridge section rather than Bridge pages that were under reference.


johannes-b (approved):
lgtm

PR #1244, opened 2022-06-01
Fix a few more xrefs

StackScribe:

Cleaning up a few more broken xrefs. 

I just deleted the section about Quick Start from the Installation page.  We are reworking all of this and what is currently there doesn't really work.


johannes-b (approved):
lgtm

PR #1250, opened 2022-06-09
restructure remediation information

StackScribe:

This PR puts existing information about remediations into the new structure:
* "remediation sequence" is a guide chapter that is part of the "Create a project" flow
* separate reference page for the remediation configuration
* separate reference page for action-provider.  And yes, this is a bit of a stretch but I wasn't sure where else to put it so it's safe here for now.

We need to enhance this information at some time but that can wait.  But please review to ensure that this content does not contain any false information.


content/docs/0.16.x/manage/remediation-sequence/index.md
```diff
@@ -0,0 +1,71 @@
+---
+title: Remediation Sequence
+description: Implement a remediation sequence
+weight: 1
+icon: setup
+---
+
```
johannes-b, line 7: Note: The evaluation sequence is not the trigger for a remediation sequence. It can be any problem that is received by Keptn. 

Can we please change the next two paragraphs from: 

```
A remediation sequence implements self-healing for a problem detected by an evaluation sequence.
It is most commonly used in production stages but can be used in any type of stage.

A remediation sequence is triggered by an event posted by an evaluation sequence.
The remediation defines the action to take to attempt to correct the problem
that the evaluation sequence identified.
```

to:

```
A remediation sequence implements self-healing for a problem detected by a monitoring solution and sent to Keptn.  The remediation defines the action to take to attempt to correct the problem.
It is most commonly used in production stages but can be used in any type of stage.
```

StackScribe: Much better!  Done!

content/docs/0.15.x/automated_operations/remediation/index.md
```diff
@@ -99,27 +99,6 @@ spec:
 
 If multiple actions are declared, Keptn sends out events in sequential order. Given the below example, the event for triggering `scaling` is sent out before the event for `featuretoggle` is fired. 
 
-```yaml
```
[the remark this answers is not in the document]
StackScribe: This was a total mistake!  The 0.15.x docs should not be modified at all by this PR!  I have restored this example.

aepfli:
i took a look at the failing build here: 

```
Non-OK status: 403 --- docs/0.16.x/reference/files/index.html --> https://docs.fileformat.com/programming/yaml/
```

This is something which we sadly can not handle, if a domain blocks us with forbidden :( - sadly there is also no possibility to add allowed response codes, eg. to just listen for 404s :( 

2 ways we can handle this:

1. we add such urls to `.htmltest.yml` to the ignore listhttps://github.com/keptn/keptn.github.io/blob/10d569b314caef5fef8ed45390aecab9121f617c/.htmltest.yml#L15-L35
2. or we deactivate the check for external URLS

StackScribe:
Either solution works for me -- whichever is fastest and easiest to implement.  This marvelous tool is only a stopgap -- by September, I expect us to be up on Docusaurus and doing HTML testing through those tools. As I am doing massive restructuring of the docs, this tool is invaluable for catching xrefs that are invalidated.

I am a bit inclined to also make this tool advisory only so it doesn't prevent merging PRs.  What do you think?

aepfli:
Imho i do think that this tool will not be superseeded by docusaurus and the tooling it provides, i do think that it will make sense to have them both ("better safe than sorry") - but this is something we can only check when docusaurus is in place.

I would simply turn off the external URL checks, and would leave this tool in place to be safe regarding regression. Even if we swap to docusaurus and doing a restructuring, the input delivered from this tool is massive when it comes to cross links. Deactivating it to be not running or always successful will get us no important feedback imho - fixing issues afterwards in a bulk is hard. I also do not think that the restructuring will be a big boom, or will it be? therefore the support of this tool will help in the overall action. 

BUT.. as i am not really the one writing documentation, i also do not think my opinion is the most important one here. In the end the tooling should help the contributors to generate concise and correctly linked documentation. With little effort to check verify that there are no errors. So if you do like to deactivate it, and the overall community thinks the same, we can for sure go ahead, and i will open a PR :)

aepfli:
i added `fileformat.com` to the list, within my last PR as i had the same issue see

https://github.com/keptn/keptn.github.io/blob/master/.htmltest.yml#L29 and https://github.com/keptn/keptn.github.io/pull/1259 

merging/rebasing master should fix the build

oleg-nenashev (approved):
The new structure looks good to me. We position auto-remediation as a key use-case, so +1 for bringing it to the top

StackScribe:
We developed some additional information about remediation in conjunction with the kilrcoda exercise so I'll use that to beef up the content.

PR #1256, opened 2022-06-09
Fix possible typo.

mnohe:


oleg-nenashev (approved):
Thanks for catching it!

PR #1259, opened 2022-06-15
Removing unnecessary documentation path

aepfli:
It looks like this reference was created for an intermediate release, and should not be part of the docs repository

image


aepfli:
caused by #1242

aepfli:
also causing 
image

oleg-nenashev (approved):

PR #1264, opened 2022-06-16
make 0.16.x show before 0.15.x

StackScribe:
This updates the weight for the 0.16.x tree so it is displayed before the 0.15.x tree


StackScribe:
I used commit -s for the original submission so I don't know why there was a DCO problem but I just went in and did the rebase and all and it seems that it is now okay.

oleg-nenashev (approved):

PR #1268, opened 2022-06-20
docs: Fix missing version flag and helm value for Git Provisioning

thisthat:
Fixes https://github.com/keptn/keptn/issues/8105
Fixes https://github.com/keptn/keptn/issues/8104

### DIFF
--- a/content/docs/0.16.x/operate/install/index.md
+++ b/content/docs/0.16.x/operate/install/index.md
@@ -234,14 +234,21 @@ Keptn consists of a **Control Plane** and an **Execution Plane**:
 
     ```console
     helm repo add keptn https://charts.keptn.sh
-    helm install keptn keptn/keptn -n keptn
+    helm repo update
+    helm install keptn keptn/keptn -n keptn --version=$KeptnVersion
     ```
+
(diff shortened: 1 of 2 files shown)


oleg-nenashev (approved):

content/docs/0.16.x/operate/install/index.md
```diff
@@ -234,15 +234,22 @@ Keptn consists of a **Control Plane** and an **Execution Plane**:
 
     ```console
     helm repo add keptn https://charts.keptn.sh
-    helm install keptn keptn/keptn -n keptn
+    helm repo update
+    helm install keptn keptn/keptn -n keptn --version=$KeptnVersion
     ```
 
+  where, `$KeptnVersion` is the version of Keptn you want to install.
```
christian-kreuzberger-dtx, line 241: should we be more specific here and mention 0.16.0?

thisthat: I would go away from encoding the version in our docs 😅 otherwise it is a nightmare to maintain. Also, in a discussion with @StackScribe, we agreed to move from docs that are version-specific. Instead, we should provide documentation for the general case and highlight differences, if any, among the different Keptn versions.

PR #1271, opened 2022-06-21
Prepare doc for 0.17

thisthat:


mowies (approved):

aepfli:
hmm it seems like we need to add some kind of handling or a process to this prepare steps (as the releases are not available during the check) maybe we should just ignore them, and hope we never have dead links?

aepfli:
i created #1272 for keeping track of this problem - as it might get lost with this pr :) and i will answer there

PR #1288, opened 2022-07-04
Add rules for naming shipyard stages

StackScribe:

https://keptn.slack.com/archives/CNRCGFU3U/p1656682769397519 points out that we do not specify rules for naming stages.  This PR solves that problem

@johannes-b @thisthat @mowies

### DIFF
--- a/content/docs/0.16.x/reference/files/shipyard/index.md
+++ b/content/docs/0.16.x/reference/files/shipyard/index.md
@@ -77,8 +77,20 @@ although most projects only use some of the constructions.
 A stage is named for the particular activity to be performed,
 such as `development`, `hardening`, `staging`, or `production`.
 Each *shipyard* file must have at least one `stage`.
-The name of the stage becomes the name of the branch in the upstream Git repository
-and the Kubernetes namespace to which services are deployed.
+The name of the stage becomes the name of the branch
+in the [upstream Git repository](../../../manage/git_upstream)
+and the Kubernetes namespace to which
+[services](../../../manage/service) are deployed.
+
+A stage can be given any meaningful name that conforms to the
+[Kubernetes Object Names and IDs](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/)
+specification, meaning:
+
+* contains at most 63 characters
+* contains only lowercase alphanumeric characters or `-`.
+Most especially, must not contain `/` or `%`.
+* starts and ends with an alphanumeric character
+
 A stage has the properties:
 
 * `name`: A unique name for the stage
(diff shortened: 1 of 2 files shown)


content/docs/0.16.x/reference/files/shipyard/index.md
```diff
@@ -77,8 +77,20 @@ although most projects only use some of the constructions.
...
+The name of the stage becomes the name of the branch
+in the [upstream Git repository](../../../manage/git_upstream)
+and the Kubernetes namespace to which
+[services](../../../manage/service) are deployed.
+
+A stage can be given any meaningful name that conforms to the
+[Kubernetes Object Names and IDs](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/)
+specification, meaning:
+
+* contains at most 63 characters
```
thisthat, line 89: Keptn reduces this value to 43: https://github.com/keptn/keptn/blob/master/installer/manifests/keptn/values.yaml#L246
However, users can change this number via the Helm value `shipyardController.config.serviceNameMaxSize`

StackScribe: Interesting, thanks!  Is there a limit to how high this value can be set -- like the k8s limit?  Are there good reasons to not increase the limit?  We may as well be thorough here, right?

thisthat: I don't recall why we selected 43 as a magic number, I'll check with the team and let you know :)

thisthat: Due to DNS limitations, Kubernetes allows only 63 char names. For the continuous delivery use case, the `helm-service` creates Helm releases, which have [53 char limits](https://helm.sh/docs/chart_template_guide/getting_started/#adding-a-simple-template-call). Keptn makes a release name with the following format: `<serviceName>-generated`. Since `-generated` is 10char long, we limit the `serviceName` part to be 43chars.

I would say that this value can increase if the user does not use the `helm-service`.

oleg-nenashev (changes requested):
Should be also duplicated to 0.17.x before merging

content/docs/0.16.x/reference/files/shipyard/index.md
StackScribe: Fabulous!  I updated the content with this new information; can update further if necessary.

StackScribe:
Yes, once we get the text set, I will copy it to the 0.17.x page

oleg-nenashev (commented):
The text looks good to me

thisthat (approved):

johannes-b (approved):
lgtm

oleg-nenashev (approved):

PR #1299, opened 2022-07-11
docs: Remove load integrations page

Kirdock:
Part of https://github.com/keptn/keptn/issues/7378
Signed-off-by: Klaus Strießnig <k.striessnig@gmail.com>


TannerGilbert (commented):
Shouldn't we remove the integrations documentation for the older versions as well because after removing https://get.keptn.sh/integrations.html in keptn/get.keptn.sh#80 getting the integrations won't work there anymore either?

TannerGilbert (approved):
The page is stored in an S3 bucket on AWS. Therefore the page is still accessible after keptn/get.keptn.sh#80.

StackScribe (approved):
lgtm

PR #1313, opened 2022-07-19
Add Keptn slack to xrefs that are ignored

StackScribe:

### DIFF
--- a/.htmltest.yml
+++ b/.htmltest.yml
@@ -19,6 +19,8 @@ IgnoreURLs:
 - "github.com/keptn/keptn/commit"
 # GitHub Help does not like GitHub Action access
 - "help.github.com"
+# Xrefs to Keptn slack are not resolving
+- "https://slack.keptn.sh" 
 # Old Announcements
 - "keptn.sh/docs/0.7"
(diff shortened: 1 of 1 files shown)


aepfli (approved):

christian-kreuzberger-dtx (approved):
LGTM

christian-kreuzberger-dtx:
FYI, I would like to see a better description in the Pull Request, or link a GH issue to it that describes the problem. I will provide an example of what I would have done below.
---

Recently during xref checks in our CI, we are getting the following error:
image
Example CI run: https://github.com/keptn/keptn.github.io/runs/7403101563?check_suite_focus=true

To fix this error, we are adding https://slack.keptn.sh into the list of things that are not to be checked.

StackScribe:
Whoops, this is not the right solution.  See https://keptn.slack.com/archives/C01E3A84PJL/p1658317863109659 .  Of course, merging this would unblock the PRs that are failing the xref CI but perhaps it is good if we have an automated test that alerts us when slack.keptn.sh is no longer working?

StackScribe:
I am merging this so we can merge other PRs while waiting for a good fix here.

PR #1314, opened 2022-07-20
fix: correct webhook-service go templating

christian-kreuzberger-dtx:
During reading the docs I found that there was a syntax error in using go-templating in webhooks.

This PR corrects the syntax error, and also slightly adapts the text (e.g., it's not about string characters, it's about alphanumeric characters. `@` is also a string, but it's not alphanumeric).


StackScribe (approved):
lgtm

PR #1315, opened 2022-07-21
Replace Prometheus/Dynatrace install instructions with link to integrations

StackScribe:
Also added Datadog to the list


arthurpitman:
@StackScribe this looks good to me from the perspective of dynatrace-service . The only issue I see is that our Table of Contents on Artifact Hub doesn't use permalinks to sub-pages but rather to the main branch, however, we can fix that there.

arthurpitman (approved):

PR #1318, opened 2022-07-26
Remove quality_gates/sli-provider

StackScribe:

Moved any interesting information from the quality_gates/sli-provider page to the sli reference page


content/docs/0.18.x/reference/files/sli/index.md
```diff
@@ -6,33 +6,43 @@ keywords: [0.18.x-quality_gates]
 ---
 
 The *sli.yaml* file contains definitions of the Service-Level Indicators (SLIs)
-defined for your Keptn installation.
+that are used to implement [quality gates](../../../../concepts/quality_gates) evaluations
```
oleg-nenashev, line 9: It is just one of the use-cases, but SLIs drive everything in Keptn.
If you remove the page, I think SLI providers should be described here too

StackScribe: Yes, this exercise is just moving the existing content into the new structure.  Then we need to do a lot of review, rewriting, and updating.  The old sli-provider page had details about Dynatrace and Prometheus and we agreed to pull those details from the main docs and have them only in the documentation associated with the Integration.  Much of the material about Dynatrace in the old monitoring section was obsolete and contradicted what is in the Dynatrace integration doc so we had discussions with Arthur Pittman and others last week and decided to delete all but high-level references from the main docs.  Other than that, all the information that was in the sli-provider doc is preserved here.

oleg-nenashev: Yes, but here you explicitly narrow down SLIs to a single use case. It wasn't like that in the text before, so for me it is rather a regression

StackScribe: It kind of was in the old text but whatever.  If I revert line 9, you'll okay it?

oleg-nenashev (approved):

PR #1322, opened 2022-07-26
Create "Update project" page

StackScribe:
This takes the content about updating a project from two different files and creates a separate file with that information.

I am not sure that these examples are that interesting or that the information is complete but this preserves the existing information into a separate file


oleg-nenashev (approved):
I think we also need a better name for this page

PR #1325, opened 2022-07-26
Create "Delete Project" and "Create Project" pages

StackScribe:

Creating separate pages for "Create a project" and "Delete a project" using the existing material.  And yes, the "Create a project" has some glaring omissions that will be rectified in a subsequent PR.

Note that I had to modify source in the reference/cli/commands directory to pass the xref check but this fix will be overwritten the next time we pull the source from keptn/keptn.  I filed https://github.com/keptn/keptn.github.io/issues/1326 .


oleg-nenashev (approved):
Fine with me

PR #1327, opened 2022-07-27
docs: Fix broken links

ghaydtner:
Fix broken links where the release part is missing in the URL.


oleg-nenashev (approved):
I wonder why the absolute URLs are used there, to start with. CC @StackScribe

StackScribe:
I don't know the history but I suspect that these older versions of the docs just used the full URLs of the referenced pages rather than the relative paths.  I confirmed with @mowies that we do not automatically update the source of the CLI reference pages for older releases so these changes will not be overwritten so am approving and merging.

Issue #1115, opened 2022-03-16
Create reference page for Keptn Cloud Events


StackScribe:
We need a reference page in Reference/Miscellaneous that lists all Keptn Cloud Events.

Currently, https://github.com/keptn/keptn/blob/master/.github/workflows/auto-update-keptn-spec.yml generates this information into keptn/spec .  This needs to be in the documentation, preferably autogenerated but with the ability to include authored content about each event, including Differences between versions information.

The documentation should include all events for  core and all services, identified by service.  For bonus points, it would be nice to enable customers to run a command that provides them with a version of this documentation that reflects only the services installed in their Keptn environment.

DavidPHirsch:
connected PR is merged

PR #1341, opened 2022-08-04
CloudEvents reference page

StackScribe:

At the 3 August 2022 Community meeting, we discussed doing reference pages that do not have full content authored in the docs but instead provide a link to the spec.  This saves users from having to prowl through the repos looking for the specs.  Our hope is that, with the new doc engine, we may be able to do something more elegant.

This is a draft of the reference page for the Keptn CloudEvents.  I am not sure what to put in the "Usage" section, but am thinking this might be a place to highlight the information that is of interest to people creating projects since the bulk of this material is mostly of use to people developing a new Keptn-service.  I am thinking that the `triggeredOn` property and its `match` sub-property require the project creator to specify an event.  What else -- other than the general understanding that is helpful for troubleshooting.

Four superfluous files added to this PR because gitlab changed their URLs and this was the PR that wouldn't pass CI because of that.

https://github.com/keptn/keptn.github.io/issues/1115


agardnerIT:
The idea of this page is excellent and as a Keptn user, there are 3 things I think I'd need to learn from this page:

1) Keptn is driven by CloudEvents
2) Keptn generates cloudevents based on sequence and task names in teh shipyard
3) The Keptn project has decided to define some common activities (such as `deployment`, `evaluation` or `test` because these are common activities of DevOps / SRE folk.
4) With point 3 in mind, the user is not limited to only those "default" events as they can choose their own task names and these are translated to cloudevents like: `sh.keptn.event.<taskname>.triggered`

agardnerIT:
Tiny typo in the `events/index.md`: Use the `TriggeredOn` property to a shipyard sequence.

1) Case is important
2) `in` not `to`

So: Use the `triggeredOn` property in a shipyard sequence.

Then it's good to go.

agardnerIT (changes requested):
Tiny typo in the `events/index.md`: `Use the TriggeredOn property to a shipyard sequence.`

Case is important so change to `triggeredOn`
`in` not `to`
So: `Use the triggeredOn property in a shipyard sequence.`

Then it's good to go.

agardnerIT (approved):

oleg-nenashev (approved):
Would be nice to get it landed

PR #1342, opened 2022-08-05
New values.yaml reference page

StackScribe:
This minimal values.yaml reference page.  In lieu of showing the file "Synopsis", this includes a link to the spec in the keptn/keptn repository.  This is a temporary measure until we determine what we can do with the new documentation engine.  Meanwhile, this spares users the effort of slogging through the source repo to find this information and we can add appropriate information as it arises plus reference it from other docs.

This draft includes a "Differences between versions" section that preserves information about changes made in 0.17.x to this file.

PR #1377, opened 2022-08-16
Revert "New values.yaml reference page"

StackScribe:
Reverts keptn/keptn.github.io#1342


StackScribe:
We fixed the problem in https://github.com/keptn/keptn.github.io/pull/1378 so this is no longer needed.

PR #1382, opened 2022-08-17
Add known limitations for Multi Cluster setup

thisthat:

Fixes https://github.com/keptn/keptn/issues/8576

### DIFF
--- a/content/docs/install/multi-cluster/index.md
+++ b/content/docs/install/multi-cluster/index.md
@@ -168,7 +168,7 @@ The important once that are used in the above example are:
 Keptn uses the integration name and version as identifier for connecting the integration to the control plane.
 Hence, in the case of multiple remote execution planes, each integration must have a different name/version pair, otherwise Keptn will assign the same set of subscriptions.
 To avoid this, we recommend to assign a different name to the integrations.
-Changing the integration name can be done by changing the Distributor `Deployment` manifest and change the environment variable `K8S_DEPLOYMENT_NAME`.
+Changing the integration name can be done by editing the Distributor `Deployment` manifest and change the environment variable `K8S_DEPLOYMENT_NAME`.
 
 ## Troubleshooting
 
--- a/content/docs/install/multi-cluster/index.md
+++ b/content/docs/install/multi-cluster/index.md
@@ -163,12 +163,13 @@ The important once that are used in the above example are:
 | `remoteControlPlane.api.token` | Keptn API token | `""` |
 
 
-## Known Limitations
-
-Keptn uses the integration name and version as identifier for connecting the integration to the control plane.
-Hence, in the case of multiple remote execution planes, each integration must have a different name/version pair, otherwise Keptn will assign the same set of subscriptions.
-To avoid this, we recommend to assign a different name to the integrations.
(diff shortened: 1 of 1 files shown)


content/docs/install/multi-cluster/index.md
```diff
@@ -163,6 +163,13 @@ The important once that are used in the above example are:
 | `remoteControlPlane.api.token` | Keptn API token | `""` |
 
 
+## Known Limitations
+
+Keptn uses the integration name and version as identifier for connecting the integration to the control plane.
+Hence, in the case of multiple remote execution planes, each integration must have a different name/version pair, otherwise Keptn will assign the same set of subscriptions.
+To avoid this, we recommend to assign a different name to the integrations.
```
StackScribe, line 170:
```suggestion
To avoid this,  you must assign a different name to each integration.
```
I'm thinking that this is not really optional -- they just have to do it, right?

content/docs/install/multi-cluster/index.md
```diff
@@ -163,6 +163,13 @@ The important once that are used in the above example are:
...
+Changing the integration name can be done by editing the Distributor `Deployment` manifest and change the environment variable `K8S_DEPLOYMENT_NAME`.
```
StackScribe, line 171: I think we need more explanation about this "Deployment manifest".  K8S_DEPLOYMENT_NAME is listed on the https://keptn.sh/docs/0.18.x/reference/miscellaneous/distributor/ reference page --  is that what you want them to modify?
That page has a section of "Configuration examples" -- maybe we should add a little section about assigning a new name for a multi-cluster environment?

content/docs/install/multi-cluster/index.md
```diff
@@ -163,6 +163,13 @@ The important once that are used in the above example are:
 | `remoteControlPlane.api.token` | Keptn API token | `""` |
 
 
+## Known Limitations
+
+Keptn uses the integration name and version as identifier for connecting the integration to the control plane.
```
StackScribe, line 168:
```suggestion
Each integration plane is assigned an identifier based on the name and version of the Keptn integration service that runs there.
Keptn uses this identifier to connect the Execution Plane to the Control Plane.
```
I'm just guessing at this wording.  The problem I see is that the rest of this section talks about the Execution Plane.  I think you are inferring that the Execution Plane identifier is based on the name of the Keptn service/integration that is installed there.  Maybe this little tidbit should be added to the bullet list at the top of the page that discusses the Execution Plane.  Just need to make the terminology consistent through the page.  It does seem that we may need to add some additional terms to the discussion.

What if one runs multiple Keptn services on an Execution Plane?  Is this identifier formed based on the name of the first service that is installed?

What happens if one runs multiple

StackScribe (changes requested):
I think this needs to be expanded a bit and, alas, I don't understand this well enough to make real good suggestions.

But actually, I am wondering why you are making this "known limitations".  I don't think this is a bug you plan to fix in the next month or two, is it?  Wouldn't it be better to just add this as a step in the "How to deploy an execution plane service?" section?  Then you can show the order -- should this be done before or after modifications to values.yaml?  And you may want to add info to the Troubleshooting section as well.  That file has a "Summary of values to configure execution plane service" -- I think those are values.yaml values, right?  So we would not be adding `K8S_DEPLOYMENT_NAME` to that table, although maybe we should specify that those are for values.yaml?

PR #1385, opened 2022-08-19
Planning your project

StackScribe:

Adam and Andi, I need a quick read and your honest opinions.  I had the idea of someone new to Keptn.  They've been through the introductory material and now are ready to create their first project, possibly for production.  It seemed that we needed a little document that supports them as they plan their first project.

So I scratched this out.  At the very best, it needs more topics and more depth -- maybe info about pros and cons of different approaches or alternate ways of accomplishing a task.  But it may be that this is the sort of piece that writers think are cool and nobody else ever reads.  I keep going back and forth so need other opinions before I do more work.  It may be that these considerations and options should be just folded into the individual pages in the "Define your project" section.  And, right now, that section just has the relevant pieces that are in the existing docs -- virtually no work has been done on them.

If you think a piece like this is worthwhile, feel free to make suggestions about additional material that should be included.  But if this is just a stupid addition, let's say so before I waste more time on it.

### DIFF
--- a/content/docs/0.19.x/manage/plan/index.md
+++ b/content/docs/0.19.x/manage/plan/index.md
@@ -28,7 +28,8 @@ after you create your project.
 A stage defines the stages and sequences of tasks that your Keptn project can orchestrate.
 The *shipyard.yaml* file you use to create a project must show the stages to be used
 – at least one stage is required but most projects have multiple stages –
-and each stage must have a name property defined. For example,:
+and each stage must have a name property defined.
+For example, a *shipyard.yaml* file like the following is enough to create the project:
 
 ```
 apiVersion: spec.keptn.sh/0.2.3
@@ -58,8 +59,8 @@ some common stage types are:
 * **testing, qa, validation** - typically used for more advanced testing
   or perhaps testing that cannot be fully automated such as usability testing.
   Your project could have multiple testing stages.
-  For example, you might have one stage for testing your Java application
-  and another stage for testing the associated database.
+  For example, you might have a `qa-java-app` stage for testing your Java application
+  and a `validate-database` stage for testing the associated database.
 * **staging, hardening** -- typically used for performance and security testing
   before the artifact is deployed into production
 * **production** -- Keptn can monitor your active production site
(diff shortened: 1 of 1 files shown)


content/docs/0.19.x/manage/plan/index.md
```diff
@@ -0,0 +1,115 @@
...
+It is important to understand that a Keptn project is **not** a pipeline.
+Instead, activities are triggered by specific events.
+For a broader explanation of this, you can read the seminal
+[Continuous Delivery without pipelines -- How it works and why you need it](https://medium.com/keptn/continuous-delivery-without-pipelines-7e84db8c8261) article.
+
+## What should this project do?
+
+Before you can run the Keptn command to create a project,
+you must populate a [shipyard.yaml](../../reference/files/shipyard) file
+that defines the `stages` for your project.
```
grabnerandi, line 24: defines stages and sequences of tasks that your keptn project will be able to orchestrate

content/docs/0.19.x/manage/plan/index.md
```diff
@@ -0,0 +1,115 @@
...
+metadata:
+  name: "shipyard-sockshop"
+spec:
+  stages:
+    - name: "dev"
+    - name: "hardening"
+    - name: "production"
+```
+
+In order to do any work, each stage must have at least one sequence and most sequences will contain one or more tasks but this information can be added to the shipyard after you create your project.
```
grabnerandi, line 42: you say here that a stage always needs to have at least one sequence - but - in the example above there are no sequences. I think its still true that there is always the "evaluation" sequence there by default (=implicit) even if you dont specify it. this is why a basic shipyard that only contain stages actually works as every stage by default has a sequence called "evaluation" with a single task that is also called evaluation. A user can however define his own evaluation sequence with any set of tasks

grabnerandi (commented):
i like this page and added some comments

content/docs/0.19.x/manage/plan/index.md
agardnerIT: Should we be “advertising” this though? I understand why it is so, but It breaks the premise that the Shipyard is the source of truth and IMO promotes bad habits.

StackScribe: I stole the statement that "Each stage must have  a sequence" from somewhere in the existing docs and then started to see that this didn't exactly make sense.  Thanks for the explanation, Andi -- now I understand.   For this section, I'm not sure we even need to mention sequences.  But I am going to add this subtle detail to the shipyard reference page.

content/docs/0.19.x/manage/plan/index.md
StackScribe: I did a bit of rewording -- see what you think.

content/docs/0.19.x/manage/plan/index.md
StackScribe: This is way too in-the-weeds for this document but I added info about the implicit evaluation sequence to the shipyard reference file in a separate PR

grabnerandi (approved):
lgtm

agardnerIT (approved):
All good from me

PR #1386, opened 2022-08-19
Explain implicit evaluation sequence in shipyard ref

StackScribe:

@grabnerandi provided technical detail about the implicit `evaluation` sequence that is provided in every stage.  Preserving that info in the shipyard reference file.

Applied to both 0.18.x and 0.19.x


grabnerandi (approved):
lgtm

PR #1387, opened 2022-08-21
Update README.md

adityyaa-10:


oleg-nenashev (changes requested):
Thanks for the suggestion! Personally I am against using HTML in Markdown just for appearance reasons

README.md
```diff
@@ -1,4 +1,4 @@
-# Keptn Website
+<h1 align="center"> Keptn Website </h1>
 
 The Keptn project has a static website powered by [Hugo](https://gohugo.io/).
 This repository contains the most of the contents though some of the resources hosted in separate repositories.
```
StackScribe, line 4:
```suggestion
This repository contains most of the contents though some of the resources are hosted in separate repositories.
```

StackScribe (changes requested):
I agree -- this seems like a gratuitous use of HTML.   What is the advantage of not using the standard markdown coding for the headers?

PR #1390, opened 2022-08-22
 Marked 0.12, 0.13, 0.14, and 0.15 as unsupported and 0.19 under development

johannes-b:
Provide a banner that shows:
* 0.12, 0.13, 0.14, 0.15 are not supported, and
* 0.19 is under development. 

Signed-off-by: Johannes Bräuer <johannes.braeuer@gmx.at>

### DIFF
--- a/content/docs/0.19.x/_index.md
+++ b/content/docs/0.19.x/_index.md
@@ -5,6 +5,7 @@ weight: 977
 sidebar_multicard: true
 icon: docs
 hide: true
+develop: This documentation is currently under development and officially released with the next Keptn release. 
 aliases:
   - /docs/0.19.0/
 ---
--- a/content/docs/0.19.x/_index.md
+++ b/content/docs/0.19.x/_index.md
@@ -5,7 +5,7 @@ weight: 977
 sidebar_multicard: true
 icon: docs
 hide: true
-develop: This documentation is currently under development and officially released with the next Keptn release. 
+develop: This documentation is currently under development for a future Keptn release.
 aliases:
   - /docs/0.19.0/
 ---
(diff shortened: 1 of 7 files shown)


johannes-b:
Hello @StackScribe , 
can you please take a look? :)

content/docs/0.19.x/_index.md
```diff
@@ -5,6 +5,7 @@ weight: 977
 sidebar_multicard: true
 icon: docs
 hide: true
+develop: This documentation is currently under development and officially released with the next Keptn release.
```
StackScribe, line 8:
```suggestion
develop: This documentation is currently under development for a future Keptn release. 
```
We can't guarantee that this will be the next release.  We have hide: true set so customers shouldn't see this anyhow.

StackScribe (approved):
So is 0.16.2-rcl going to be supported?  Do we have Release Notes that explains the situation?  Do we need an upgrade guide for people going from 0.15.x to 0.18.x?

I am approving to keep things moving forward -- I assume that you know what you are doing here.

oleg-nenashev (approved):

oleg-nenashev:
> So is 0.16.2-rcl going to be supported? Do we have Release Notes that explains the situation?

Good question for @johannes-b 

> Do we need an upgrade guide for people going from 0.15.x to 0.18.x?

Nope

content/docs/0.19.x/_index.md
johannes-b: Fixed it. Thx!

johannes-b:
@StackScribe and @oleg-nenashev , many thanks for raising concerns here! 

This gave me a trigger to re-think the situation and we should not mark 0.16, and 0.17 as unsupported. Instead, I added the banner to 0.12 - 0.15 since deep links might still lead the user to these versions.

PR #1400, opened 2022-08-26
link to Adam's "Trigger a sequence" video from "Triggers" page

StackScribe:

### DIFF
--- a/content/docs/0.19.x/define/triggers/index.md
+++ b/content/docs/0.19.x/define/triggers/index.md
@@ -16,7 +16,7 @@ A Keptn sequence can be triggered in any of the following ways:
 ## Trigger a sequence from the Keptn Bridge
 
 The [Trigger a Keptn sequence](https://www.youtube.com/watch?v=S0eumPKuAJY) video
-ilustrates how to trigger a sequence using the Keptn Bridge
+illustrates how to trigger a sequence using the Keptn Bridge
 and discusses using `triggeredOn` in a `sequence` in your [shipyard](../../../0.19.x/reference/files/shipyard).
 
 ## Use triggeredOn in a sequence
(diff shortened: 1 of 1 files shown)


agardnerIT (commented):
Tiny typo: `ilustrates`

StackScribe:
> Tiny typo: `ilustrates`

Fixed

agardnerIT:
When `triggeredOn` is used, the shipyard author is implicitly linking two sequences, tehrefore it probably wouldn't make sense to manually trigger the production sequence if it also has a `triggeredOn` block. Think of it like a Jenkins pipeline. If your production release pipeline triggers only when hardening is successful, that's fine. I can't imagine why you'd skip the UAT safeguards and instead manually trigger a production release.

agardnerIT:
One final change and it's good to go: `If the delivery sequence fails because of a failed test task`. Reword that to `IF the delivery sequence fails`.

If any task is a sequence fails, the sequence fails and thus it would match your `fail` selector.

agardnerIT (changes requested):
When `triggeredOn` is used, the shipyard author is implicitly linking two sequences, therefore it probably wouldn't make sense to manually trigger the production sequence if it also has a `triggeredOn` block. Think of it like a Jenkins pipeline. If your production release pipeline triggers only when hardening is successful, that's fine. I can't imagine why you'd skip the UAT safeguards and instead manually trigger a production release.

All that's to say, perhaps remove the section about HTTP `POST`ing to trigger the production sequence as it has a `triggeredOn` condition.

One final change and it's good to go: If the delivery sequence fails because of a failed test task. Reword that to `If the delivery sequence fails.`

If any task is a sequence fails, the sequence fails and thus it would match your fail selector.

StackScribe:
I fixed the "If the delivery sequence fails..." sentence.  Are you okay with leaving the HTTP POST section alone for now?   I'm thinking that we have lots of problems with lots of the examples in what is now the "Define" section -- we can revisit everything as we try to put this together.

agardnerIT (approved):
LGTM

PR #1403, opened 2022-08-27
Move service page to "Manage" section

StackScribe:

* Move the "Create service" page to "Manage" section
* Add instructions for creating a service from the Keptn Bridge.  I will link Adam's "Create project and service video" to that section after it is moved to the Keptn youtube area.  I want to get this PR merged quickly so I can reference it from other docs.
* A few minor edits

### DIFF
--- a/content/docs/0.19.x/define/delivery_sequence/index.md
+++ b/content/docs/0.19.x/define/delivery_sequence/index.md
@@ -121,7 +121,7 @@ After declaring the delivery for a project in a shipyard, you are ready to creat
 
 ## Trigger a multi-stage delivery
 
-After creating a project and creating your [service(s)](../service), you can trigger a delivery using the Keptn CLI or API.
+After creating a project and creating your [service(s)](../../manage/service), you can trigger a delivery using the Keptn CLI or API.
 
 <details><summary>**Trigger via Keptn CLI**</summary>
 <p>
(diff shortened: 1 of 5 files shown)


agardnerIT (changes requested):
The top of part of the page looks good but the helm chart instructions are irrelevant here. You can create a stage and not have a Helm chart. What if your sequence doesn't involve delivery but other tasks that have nothing to do with helm?

I'd remove the bottom section and put it elsewhere. Good info but I think it'll confuse because I read that page as "I create a service and now I need a helm chart, regardless of what I want to do".

StackScribe:
That part about the helm-chart requirements is legacy -- I just made it a header.  Where else would you put this information?  I'm thinking that maybe it belongs here but needs to be worded differently.  Maybe begin with "If your service does xxx or yyy, you need to provide a helm chart to..."  This section should be comprehensive and tell the reader everything they need to do to get their project up and running, including in a production environment, and this is sometimes part of that, right?

StackScribe:
I made this a draft until I can fix the xrefs

StackScribe:
xrefs are fixed

agardnerIT:
> That part about the helm-chart requirements is legacy -- I just made it a header. Where else would you put this information? I'm thinking that maybe it belongs here but needs to be worded differently. Maybe begin with "If your service does xxx or yyy, you need to provide a helm chart to..." This section should be comprehensive and tell the reader everything they need to do to get their project up and running, including in a production environment, and this is sometimes part of that, right?

Yes, if your intention is to orchestrate the delivery of a helm chart using Keptn, you probably want a task called `deployment` ([as per the standardised events in the spec](https://github.com/keptn/spec/blob/master/cloudevents.md#deployment)) then you need to bring a tool (eg. job executor service) to action that deployment and finally you'll need to add the helm chart to the Keptn upstream repo.

agardnerIT (approved):
Approving as this is a file movement rather than a broad rewrite. We can move and file separate PRs for any identified improvements.

PR #1405, opened 2022-08-28
Planning: Tool calls Keptn or Keptn calls Jenkins

StackScribe:

This captures a response @agardnerIT gave in https://keptn.slack.com/archives/CNRCGFU3U/p1661478530360699 .

This is useful information that is not otherwise included in the documentation.  Feel free to refine the text

### DIFF
--- a/content/docs/0.19.x/manage/plan/index.md
+++ b/content/docs/0.19.x/manage/plan/index.md
@@ -79,6 +79,49 @@ You must identify each tool you use to Keptn;
 this can be done in a variety of ways.
 See [Keptn and other tools](../../../concepts/keptn-tools) for more information.
 
+## Who calls whom?
+
+When integrating certain tools with Keptn,
+you have a choice of having your tool call Keptn
+or having Keptn call your tool.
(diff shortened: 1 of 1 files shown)


agardnerIT:
1. I would flip the sentences around to have the current method (and code snippet) first. The shared library is only applicable to Keptn versions 0.12.x and older so that should go further down the page.

2. The current "best practice" is for tools (Jenkins) to be triggered from Keptn sequences (so Jenkins would respond to the task event). Wording to that effect should go first. If a user really does want to integrate "the other way around" (for example to trigger only a quality gate from Jenkins) they still need a Keptn sequence with a single `evaluation` task. In that respect, they don't "gain" anything and, it could be argued, actually lose something because they need to code logic into their pipeline to wait for the evaluation result. All-in-all it's a harder, more legacy approach.

oleg-nenashev (approved):
The text looks good to me

StackScribe:
Thanks, Adam.  I rewrote as you suggested and made the prose a bit more opinionated about which approach is preferred.

PR #1407, opened 2022-08-29
Start a project

StackScribe:

Simple instructions to start a project using either the Keptn Bridge or the command line.

### DIFF
--- a/content/docs/0.19.x/manage/git_upstream/index.md
+++ b/content/docs/0.19.x/manage/git_upstream/index.md
@@ -1,14 +1,15 @@
 ---
 title: Git-based upstream  
-description: Select a Git-based upstream for a project.
-weight: 20
+description: Create a Git-based upstream for a project.
+weight: 30
 keywords: [0.8.x-manage]
 aliases:
   - /docs/0.8.0/manage/git_upstream/
   - /docs/0.8.1/manage/git_upstream/
 ---
 
-Keptn manages a project in an internal Git repository which is separate from the repository that stores your source code.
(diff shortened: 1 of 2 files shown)


agardnerIT:
`specified to that command` >> `required for that command` perhaps?

This shipyard would break the webhook screen in the bridge and cause an invalid shipyard warning because it doesn't contain `sequences` and corresponding `tasks` blocks for each stage.

It's fine to show it like this as long as we immediately link to some information telling users that after initial project creation, they must then decide on sequences and tasks. I like that flow because it reinforces that stages are static but sequences and tasks can be changed at will.


```
apiVersion: spec.keptn.sh/0.2.3
kind: "Shipyard"
metadata:
  name: "shipyard-sockshop"
spec:
  stages:
    - name: "dev"
    - name: "hardening"
    - name: "production"
```

StackScribe:
> `specified to that command` >> `required for that command` perhaps?

done

StackScribe:
added rules for project name; put the configuration parameter in the shipyard reference page -- only need that in one place and it isn't when they are learning how to create a project.

Not sure what to do about the config parameter for service name max length -- I don't think a service name is specified in shipyard and I don't have a reference page for services -- probably need one.

StackScribe:
Added link to "Define Keptn Projects" for information about how to code sequences and tasks.  Right now, that section is just a dumping ground but hopefully I will improve the structure very soon.

oleg-nenashev (approved):
Nice improvement. I am concerned we store the videos outside Keptn channels

PR #1412, opened 2022-08-31
fix: Added aliases for version check documentation starting 0.14.x

### LINKED ISSUE #1411: Keptn Bridge version check documentation link broken since 0.14.x
In #1167 the version check documentation was moved from `content/docs/0.14.x/reference/version_check/index.md` to `content/docs/0.14.x/bridge/version_check/index.md` which broke the link displayed in the Keptn Bridge.

bridge_version_check

There are multiple ways of fixing the described issue:

1. Update the link inside the bridge. This is not a good option because if only the master is changed, customers with versions between `0.14` and the current version `0.18.2` would still have the problem.
2. Move the `version_check` back into the `reference` folder
3. Redirect `https://keptn.sh/docs/<version>/reference/version_check/` starting from `0.14.x` to `https://keptn.sh/docs/<version>/bridge/version_check/`

TannerGilbert:
## This PR

- Fixes broken link on Keptn Bridge starting Keptn 0.14.x by adding an alias to version check documentation.

### Related Issues

Resolves #1411

Signed-off-by: TannerGilbert <gilberttanner.work@gmail.com>

### DIFF
--- a/content/docs/0.14.x/bridge/version_check/index.md
+++ b/content/docs/0.14.x/bridge/version_check/index.md
@@ -2,6 +2,8 @@
 title: Daily Version Check
 description: Enable/Disable daily version check feature
 weight: 40
+aliases:
+  - /docs/0.14.x/reference/version_check/
 ---
 
 The feature of a daily version check informs a Keptn user in the Keptn Bridge about:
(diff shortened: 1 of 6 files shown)


thisthat (approved):
LGTM

heinzburgstaller (approved):

PR #1424, opened 2022-09-06
Remove outdated FAQ reference to keptn install on Quickstart page

agardnerIT:
This PR:

- Removes an FAQ entity because it refers to the outdated `keptn install` method

Signed-off-by: agardnerit <adam@agardner.net>

### DIFF
--- a/content/docs/quickstart/_index.md
+++ b/content/docs/quickstart/_index.md
@@ -257,7 +257,7 @@ k3d cluster delete mykeptn
 
 ### Disk pressure on pods in Kubernetes
 
-If the installation of Keptn is timing out, you can check if the root cause are low resources, such as disk space, by executing `kubectl describe pod PODID -n keptn`.
+If the installation of Keptn is timing out, you can check whether this is caused by low resources such as disk space, by executing `kubectl describe pod PODID -n keptn`.
 
 Please ensure that your local k3d environment has enough resources. You can verify this in your Docker resource settings. This quickstart guide has been tested with the following configuration:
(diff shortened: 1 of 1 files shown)


content/docs/quickstart/_index.md
```diff
@@ -255,20 +255,9 @@ k3d cluster delete mykeptn
...
-helm install helm-service https://github.com/keptn/keptn/releases/download/0.14.1/helm-service-0.14.1.tgz -n keptn --create-namespace --wait
-helm install jmeter-service https://github.com/keptn/keptn/releases/download/0.14.1/jmeter-service-0.14.1.tgz -n keptn --create-namespace --wait
-```
-
-Now continue with step 4 from the quickstart guide.
-
 ### Disk pressure on pods in Kubernetes
 
-If the installation of Keptn is timing out, you can check if the root cause are low resources, such as disk space, by executing `kubectl describe pod PODID -n keptn`.executing
+If the installation of Keptn is timing out, you can check if the root cause are low resources, such as disk space, by executing `kubectl describe pod PODID -n keptn`.
```
StackScribe, line 260:
```suggestion
If the installation of Keptn is timing out, you can whether this is caused by low resources, such as disk space, by executing `kubectl describe pod PODID -n keptn`.
```
The "root cause are low resources" construct is problematic -- I think it should be "is" rather than "are" to match the singular "root cause" but I suggest rewording to just eliminate the problem.

StackScribe (approved):
One tiny edit suggestion but this looks good.

Issue #1396, opened 2022-08-25
How to point to different sli-provider environments for different stages, projects


StackScribe:
Info from @grabnerandi  in https://keptn.slack.com/archives/CNRCGFU3U/p1661352188359829 .

agardnerIT:
Does this duplicate https://github.com/keptn/keptn.github.io/issues/1298 ?

PR #1425, opened 2022-09-06
How to use different SLI environments for different projects/services/stages

StackScribe:
This preserves information @grabnerandi provided to a customer question.

I tried to genericize the info while using Dynatrace as an example -- I couldn't figure out where this is in the Prometheus configuration.  And I'm not sure what I wrote is totally accurate and clear.

We probably need an example of this but maybe we should build it into a bigger sample in the "define" guide section and just link to that here.

https://github.com/keptn/keptn.github.io/issues/1396

### DIFF
--- a/content/docs/0.19.x/reference/files/sli/index.md
+++ b/content/docs/0.19.x/reference/files/sli/index.md
@@ -75,7 +75,7 @@ use the [keptn add-resource](../../cli/commands/keptn_add-resource) command:
 
 **Example of multiple SLI configurations:**
 
-* Let's assume, we add the following SLI configurations to a project, stage, and service: 
+* Let's assume we add the following SLI configurations to a project, stage, and service: 
 
     * SLI configuration on project-level:
 
@@ -118,7 +118,7 @@ use the [keptn add-resource](../../cli/commands/keptn_add-resource) command:
     ```
 ## Using different SLI environments
 
-Different projects, stages, and services can utilize different environments for your SLI.
+Different projects, stages, and services within stages can utilize different environments for your SLI.
 To do this, you create different configuration files for your SLI
 and have each point to a unique secret that contains the API endpoint for your SLI
 and a token.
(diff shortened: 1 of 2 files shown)


agardnerIT (changes requested):
`dynatrace.conf_dev.yaml` is not valid. The filename and folder path must always be `dynatrace/dynatrace.conf.yaml`. What changes is where you place that file in Git. Imagine a project with three stages: `dev`, `qa` and `prod`. Keptn will create 3 Git branches: `dev`, `qa` and `prod` so you could have 3x `dynatrace.conf.yaml` files, one on each branch.

StackScribe:
https://keptn.slack.com/archives/D033ABY5LKC/p1662896408763279 includes the following from @grabnerandi : Hi. It seems Adam has already given some feedback on what you wrote in that PR. He is correct that the filename for the dynatrace service always has to be dynatrace.conf.yaml - BUT - the dynatrace-service can either load this file for a specific service&stage, for a whole stage or for the whole project. As a user you can therefore upload different versions of dynatrace.conf.yaml to either the project, the stage or for a service in a stage. This is also documented here in more detail --> [https://github.com/keptn-contrib/dynatrace-service/blob/master/documentation/dynatrace-conf-yaml-file.md#customizing-the-conf[…]ptn-stage-or-service](https://github.com/keptn-contrib/dynatrace-service/blob/master/documentation/dynatrace-conf-yaml-file.md#customizing-the-configuration-for-a-specific-keptn-stage-or-service)
For the prometheus service i think this is not implemented the same way. for prometheus "I THINK" you would need to install different preometheus-services and let each point to a different prometheus instance. But - I am not the expert here - so - you may want to ask somebody that is more famliar with the prometheus-service implementaton

StackScribe:
I think I have corrected the information as you recommended.  I also made a few small changes to the concepts/quality_gates page to mention this capability and point to the reference page for details.

oleg-nenashev (approved):
LGTM

agardnerIT (approved):

PR #1451, opened 2022-10-11
Proposal: Canonical Approach with latest branch

aepfli:
This PR changes the current version docs to latest and adds canonical links to the old docs (where it is easily possible).

Instead of copying the folder and name it for the new version the process would look like:
1. copying latest for the old version
2. change the alias in index.md of latest to the new version

All links will stay the same, the only issue which might arise is the CLI update, which we would need to investigate specially.

Also this allows for easier version specific changes/updates.

:information_source: this is just a proposal currently - it would fix some back link issues, it is not perfect, but the canonical URLs, will point to the latest version all the time, therefore SEO should always point to the latest release (at least where we can map it easily, not working for moved pages)


aepfli:
seems like my page exists check is broken, I am gonna fix it :)

StackScribe:
lgtm but @mowies should review it.   Also note that, for LTS, we are going to move the Concepts, Installation, etc sections under LTS or 1.x or something so they will all be at the top level, which means that we probably do not need to redo all the xrefs to use "latest" rather than the actual release number.  But having latest in the URL gives us a lot more flexibility for the future and may even mean that we can silently roll the LTS docs into the new doc engine without breaking user bookmarks and such.

oleg-nenashev (commented):
I think it collides with the `v1` approach we discussed. I would recommend using it as "latest" for LTS, and introduce another "latest" branch only after we have substantial changes in the documentation

aepfli:
https://developers.google.com/search/docs/crawling-indexing/consolidate-duplicate-urls is a good read from google regarding canonical URLs

Issue #1472, opened 2022-10-17
Remove references to KEP-70


StackScribe:
We have links to KEP-70 throughout the docs when we mention that stages cannot be added, removed, or renamed in a project after the project is created.  Since this project is no longer active, references to it should be removed.

PR #1477, opened 2022-10-18
Remove references to KEP-70

StackScribe:

KEP-70 to allow adding/deleting/modifying stages in a shipyard after it is created is no longer active so deleting all references to it from the docs.

https://github.com/keptn/keptn.github.io/issues/1472


oleg-nenashev (approved):
I am planning to merge the PR and to add a Draft&Stalled behavior. After that a link could be just updated, but I agree it is not necessary in docs. +1 for merge

PR #1480, opened 2022-10-23
Add Carousel with the key events to the website

oleg-nenashev:
## Tasks

- [x] Add the carousel engine, based on patched Owl Carousel compatible with Hugo 0.53 
- [x] Add Kubecon, Keptn LTS, and Keptn Lifecycle Controller as entries
- [ ] Add graphics


thisthat (approved):

AloisReitbauer (commented):
LIfecycle controller is not a sub project

oleg-nenashev:
> LIfecycle controller is not a sub project

It is. We agreed it would be called like that during the hosting process and the community meetings. All the discussions were public, and I believe there was a consensus. Sub-project structure/approach is a part of the Keptn documentation and a common approach in the CNCF, and w. If you disagree with this naming, I suggest to start a GitHub Discussion so that we could find better naming

References:

* Community Membership doc defining the subproject framework: https://github.com/keptn/community/blob/main/COMMUNITY_MEMBERSHIP.md
* https://github.com/keptn-sandbox/contributing/issues/18
* Experimental project under the repository: "It is an experimental project, under the umbrella of the [Keptn Application Lifecycle working group](https://github.com/keptn/wg-app-lifecycle)." in https://github.com/keptn/lifecycle-controller

PR #1482, opened 2022-10-24
Landing banner updates

oleg-nenashev:
I got a request to remove the carousel in favor of the news list header, and hence I implemented so. LGTM for me on the desktop, narrow and mobile screens. Keeping the carousel code for the future

image


AloisReitbauer (changes requested):
Can we move the partial announcement block back to the top. Also blue on blue is very hard to read.

oleg-nenashev:
> Can we move the partial announcement block back to the top.

@AloisReitbauer  It requires more changes in Javascript and CSS. The website does not use standard controls for that from Hugo templates, and there is a collision between the header and the announcement layout once you make the text longer than one line. May need a few hours to get it sorted, will try to do it today

>  Also blue on blue is very hard to read.

Can do. We just use such rendering for links elsewhere too

layouts/partials/announcement.html
```diff
@@ -1,8 +1,13 @@
...
   <div class="container">
     <div class="row">
-      <div class="col-12 d-flex flex-column text-center">
-        <h3>Keptn 0.19.1 is available! Find all <a href="./docs/news/release_announcements/keptn-0191/">details on this release here</a>. </h3>
+      <div class="col-12 d-flex flex-column box-announcements">
+        <ul>
+          <li>Join us for Kubecon and <a href="https://community.cncf.io/events/details/cncf-keptn-community-presents-keptn-community-day-kubecon-cloudnativecon-detroit/">Keptn Community Day</a> in Detroit!
+              Community booth in the project pavillion, cool demos, main track talks, and a first ever full-day Keptn event.
+          <li>Check out the <a href="https://github.com/keptn/lifecycle-controller">Lifecycle Controller</a>! Cloud native pre-, post-deployment and application health checks.
+          <li>Keptn 0.19.1 is available! Find all <a href="./docs/news/release_announcements/keptn-0191/">details on this release here</a>.
```
StackScribe, line 9:
```suggestion
          <li>Keptn 0.19.1 is available! Find <a href="./docs/news/release_announcements/keptn-0191/">details here</a>.
```

StackScribe (approved):
I like it!  I made one editorial change for your consideration.

PR #1504, opened 2022-11-03
Upgrading hugo to 0.105

aepfli:
Upgrading to a more recent version of hugo.
The page is building, and looks good, but there might be hidden regressions which somebody working with the documentation should actually inspect

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -133,7 +133,7 @@ The makefile targets are utilizing docker for easier setup. This requires no add
 
 ### Using a locally installed Hugo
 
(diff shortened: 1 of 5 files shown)


StackScribe (commented):
The error message I've been getting is to replace `Page.Dir` with `File.Dir`.  I see `Page.File.Dir` here -- is this correct?

We also need to provide instructions for people to upgrade their private versions of Hugo.  That can be a separate PR.  If we can figure out how to get me updated, I'll then fix the CONTRIBUTING.md file.

Thanks so much!

aepfli:
it seems like this changes are also working with the old version of hugo, so an information in CONTRIBUTING.md should be enough to point out the new version. Anyways there is the makefile, which offers a good an automated entrypoint for everyone.

thisthat (approved):
From the preview build, everything looks correct to me! Thank you @aepfli 🙌

Issue #1511, opened 2022-11-08
Use 1.0.0 in all command lines, etc


StackScribe:
Need to go through all 1.0 docs and update more references to 0.19.1 to be 1.0.0.  Most of these are in sample command lines but need to check them all.

Note that this is separate from https://github.com/keptn/keptn.github.io/issues/1483 , which is a final check when we publish the 1.0.0 docs.

StackScribe:
Done

PR #1515, opened 2022-11-09
Use 1.0.0 for command and file examples -- rest of files

StackScribe:

I think this completes this effort but will check after everything is merged

Part of https://github.com/keptn/keptn.github.io/issues/1511


content/docs/1.0.x/manage/tasks/index.md
```diff
@@ -156,7 +156,7 @@ The response will be a JSON object in the following format:
          "inputProperties":{
             "configurationChange":{
                "values":{
-                  "image":"docker.io/keptnexamples/carts:0.13.2"
+                  "image":"docker.io/keptnexamples/carts:1.0.0"
```
mowies, line 159: This doesn't need to be replaced, there's no carts example with 1.0.0 version

mowies (changes requested):

content/docs/1.0.x/manage/tasks/index.md
StackScribe: We have carts and sockshop all through the docs and I'm not sure what to do about it.  We could go through and replace examples with podtatohead examples or even some generic `myapp` string.  But it sounds like we are out to kill Keptn 1.0 ASAP so is it worth it

Meanwhile, what do we do with this line?  It's an example of the JSON object returned.  It doesn't make sense to leave it as 0.13.x or even 0.19.x (I don't think carts works for 0.19.x either...

StackScribe: Okay, I switched this file to use "myapp" rather than "sockshop" and "some-service" rather than "carts".  And we can discuss later whether we want to spend the time to do this throughout the docs

thisthat (approved):

mowies (approved):

Issue #1522, opened 2022-11-15
How to access a vairable when triggering an event


StackScribe:
https://keptn.slack.com/archives/CNRCGFU3U/p1668459870744349

agardnerIT:
The linked PR was closed so I'm closing this.

PR #1524, opened 2022-11-16
How to send variable in a sequence

StackScribe:

Captures info from a #help exchange

https://github.com/keptn/keptn.github.io/issues/1522


StackScribe:
@grabnerandi sent the following in Slack: "I think its not worth adding this because the example was very specific to the user and their webhook. we are referring to specific tools they are using and i think that confuses people
[3:15](https://keptn.slack.com/archives/D033ABY5LKC/p1668683729481399)
sorry that you put that much effort in this - but - i suggest we dont add this as it would probably add more confusion unless we really find time to create a more generic descripton of this"

PR #1527, opened 2022-11-22
[Proposal] Exploration with Docsy

aepfli:
Docsy is a widely used documentation theme, and it offers a lot of features we are looking for, specially versioning etc.

Sadly it still does not offer multipage support, but at least we are heading into the right direction, and it seems like an easy fix for a lot of our problems.


aepfli:
it does not look really appealling but before I put more effort into this theme, I want to challenge it with @StackScribe @DavidPHirsch and @mowies, @thschue 

the default hugo theme we are using, seems to be not the ideal candidate for documentation docsy on the other hand is widely adopted, has integrated search feature for google search or others, offers a wide variety of easily adding links etc.

Overall there would be a need to take a closer look at what we really need in the future, and if we want to remove some things from the page after all.

mowies:
What happened to the docusaurus GSoC project?

DavidPHirsch:
would add that we should keep with the recommended approach from CNCF https://github.com/cncf/techdocs/blob/main/docs/services.md#netlify--website-setup

aepfli:
> What happened to the docusaurus GSoC project?

the question is, how much effort do we want to put into the documentation moving forward? the docusaurus project is not in a ready state to be deployed, there is still efforts todo, and we do not have that much expertise with it. Hugo has proven to be solid and easy to use, but the theme we are using is lacking certain aspects. i stumbled upon this, during my research and I gave it a short try. The documentation part works great and as intended, the landing pages would need some effort.

I did this during the exploration for KLT, as there is also the discussion what to use for KLT, and ideally we keep it as simple as possible with little changes to the keptn universe. With the lowest efforts, this would mean we use Hugo as it is the current proven way. - but this is up for dicussion

aepfli:
if we want to move forward, i would suggest following approach:
1. fixing the landing pages etc. so they are in a well suited way
2. create a branch for each version, which will be displayed, and changing the hierarchy to not contain a version
3. setup multiversions builds based on those branches, and create either subdomains or any kind of other approach (we need to check on this, but a lot of tools are using this documentation so we can maybe get a solution from them)

aepfli:
it even support code file includes, etc - https://www.docsy.dev/docs/adding-content/shortcodes/#include-code-files

thschue:
I like docsy, it makes a clear impression on me, and as long as it supports what we need, I'm fine with it. @DavidPHirsch is there another CNCF recommended way for building doc sites? As far as I've seen, docsy is used by some CNCF projects (flux, opentelemetry and last but not least kubernetes)

StackScribe:
@mowies As @aepfli said, the GSoC Docusaurus project did not leave us ready to deploy and we don't have resources right now to move forward with that but it is not off the table.  The biggest question is whether Docusaurus and markdown can give us the multi-repo support that we require or if we need to go to something like Antora/asciidoc.  So we will release the Keptn LTS docs using Hugo and are looking for small improvements we can make there.    We may use a different doc engine for the lifecycle-controller docs and, if they work out well and Keptn survives, we can always migrate the Keptn docs at a later time.  Meanwhile, the OpenFeature docs are using Docusaurus so we'll get some experience with Docusaurus there.  So everything is still on the table, as it were, for the future.  And if you have opinions and/or advice, we welcome it!

DavidPHirsch:
@aepfli @StackScribe @mowies @thschue can we close this? there hasn't been any activity for some time

mowies:
Yeah I think we can close it

PR #1530, opened 2022-11-30
SEO Improvements

aepfli:
Adding canonical links to all pages and handling those missing links

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -21,6 +21,7 @@ ignoreFiles = [
 
 [params]
     google_analytics_id="UA-133584243-1"
+    version="simonsTest"
 
 [params.carouselHomepage]
     enable = true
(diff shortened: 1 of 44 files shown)


config.toml
```diff
@@ -21,6 +21,7 @@ ignoreFiles = [
 
 [params]
     google_analytics_id="UA-133584243-1"
+    version="0.19.x"
```
aepfli, line 24: this is the current version used for the canonical URLs, as soon as 1.0.x is live we need to adapt this string

StackScribe: Cool.  Let me note this down -- when i do the PR that releases the 1.0 docs, I'll include this.

StackScribe (approved):
lgtm -- thanks for doing this!

PR #1534, opened 2022-12-02
Remove ruxit agent

aepfli:
The ruxit agent configuration was outdated, the host it tried to connect to was unreachable. Althought this data might be interesting, we also do have google analytics and scarf to also generate this information.

If we want to enable the ruxit agent again, we need to think about which host we are using for reporting.


aepfli:
@christian-kreuzberger-dtx maybe you also have some input regarding this topic :) as it seems like you had discussions in the past about this topic.

christian-kreuzberger-dtx (approved):
I had no idea we had that (was introduced in https://github.com/keptn/keptn.github.io/commit/f7d95901b529f829e4bcdd0577d703ec7a3a7231 ) before I was even part of the project.

I agree with removing it.

grabnerandi:
@christian-kreuzberger-dtx - i found a conversation we had on our internal slack almost 3 years ago about it :-) - i also forgot. I think this was probably Juergen in the early days adding some dynatrace monitoring.

christian-kreuzberger-dtx:
> @christian-kreuzberger-dtx - i found a conversation we had on our internal slack almost 3 years ago about it :-) - i also forgot. I think this was probably Juergen in the early days adding some dynatrace monitoring.

Just found the same discussion :D
Pretty sure this was something that was eventually delegated to Juergen or Daniel.

PR #1538, opened 2022-12-05
add link rel preconnect to improve page load times

ghaydtner:
Add preconnect to the most used/important domains to speed up page loading.


ghaydtner:
Bug in htmltest breaks the build. I've reported the bug and will close that for now:
https://github.com/wjdp/htmltest/issues/202

PR #1549, opened 2022-12-15
Add link to KLT in the menu

thisthat:

Preview: https://deploy-preview-1549--keptn.netlify.app/

### DIFF
--- a/content/docs/_index.md
+++ b/content/docs/_index.md
@@ -1,6 +1,6 @@
 ---
-title: Keptn v1 (LTS) Docs
-linktitle: Keptn v1 (LTS) Docs
+title: Keptn v1 Docs
+linktitle: Keptn v1 Docs
 description: Learn how to use Keptn.
 sidebar_multicard: true
 icon: docs
--- a/content/docs/_index.md
+++ b/content/docs/_index.md
@@ -1,6 +1,6 @@
 ---
-title: Keptn LTS Docs
-linktitle: Keptn LTS Docs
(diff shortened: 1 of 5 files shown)


StackScribe (approved):
I am approving but I see a couple problems here.  I defer to your judgement whether to hold the release until they can be fixed:
* On the landing page, we have "Keptn Lifecycle Controller" and "Docs" -- the latter links to the Keptn docs.  Maybe just say "Keptn Docs" or "Keptn LTS Docs"?  We need more real estate on that line, I realize.  I'd lose the "Integrations" link -- that is Keptn only and is linked in the Keptn docs and it's less significant now anyhow.  Maybe the "Quick Start" should also be removed -- it is Keptn only and is in the docs.  The "Resources" tab doesn't look terribly valuable to me right now, either -- I wonder when anyone last updated it.
* Once I go into the Keptn Lifecycle Toolkit, I have no path back to the top-level page.  That is also true of the "Tutorials" page so maybe there is no solution.  But when I go into the Keptn docs, I still have the link to "Keptn Lifecycle Controller" on the top menu along with all the others.  And note that, when I go into the KLT->Docs section, I still have a "Docs" link at the top that is for the KLT docs.  So a bit confusing all around.

thisthat:
Once we moved to the new doc engine, we could solve the link problem like Flux did in their doc: https://fluxcd.io/flux/
If you press on "Docs" you can select which project you want the docs for. I like the idea of changing the current "Docs" to "Keptn LTS Docs" to avoid confusion 👍

PR #1555, opened 2022-12-22
Add aliases for moved pages

aepfli:
within 0.19 we added aliases for moved pages - we have forgotten to also move them to the new version. This is fixing the build for the future.


grabnerandi (approved):
lgtm

PR #1560, opened 2023-01-10
[PROPOSAL] Replacing hardcoded version with param

aepfli:
As we do now have a param to get the latest version, we can use this wihin our concepts, install, and explore to link to the most up-to-date documentation part. Instead of a hardcoded version which needs to be changed regularly. As a result we do have a lot of links pointing to outdated versions.


aepfli:
maybe we do not want to have this for concepts and install - but explore might be still crucial

content/docs/concepts/delivery/index.md
```diff
@@ -11,7 +11,7 @@ Keptn allows you to declaratively define *multi-stage delivery* workflows by def
...
+The definition is manifested in the [shipyard](../1.0/reference/files/shipyard] file that defines a task sequence for delivery. It can hold multiple stages, each with a dedicated *deployment strategy*, *test strategy*, as well as a *remediation strategy*. Following this declarative approach, there is no need to write imperative pipeline code. Keptn takes the shipyard file and creates a multi-stage workflow each stage having a deployment strategy (e.g., blue/green), testing strategy (e.g., functional tests or performance tests), and an optional automated remediation strategy for triggering self-healing actions. To learn more about a shipyard configuration, please continue [here](../../{{< param "version" >}}/continuous_delivery/multi_stage/).
```
StackScribe, line 14:
```suggestion
The definition is manifested in the [shipyard](../1.0/reference/files/shipyard] file that defines a task sequence for delivery. It can hold multiple stages, each with a dedicated *deployment strategy*, *test strategy*, as well as a *remediation strategy*. Following this declarative approach, there is no need to write imperative pipeline code. Keptn takes the shipyard file and creates a multi-stage workflow each stage having a deployment strategy (e.g., blue/green), testing strategy (e.g., functional tests or performance tests), and an optional automated remediation strategy for triggering self-healing actions. 
```
All this info about continuous delivery needs to be purged and I haven't gotten to it.  But we can at least remove this statement -- the shipyard reference page (referenced in the first sentence of this section) is the most comprehensive summary of shipyard.

StackScribe (commented):
Interesting.  I need to understand exactly how this works.  As a user, what determines the value of "version" for me?  It is still really ugly that the 0.x.x version of "Concepts" and "Installation' show at the top level of the docs, although perhaps it is less fatal now that one lands in the 1.0.x docs when one opens keptn.sh...

Let's think this through -- it may indeed be a very clever implementation...

aepfli:
> 
>Interesting. I need to understand exactly how this works. As a user, what determines the value of "version" for me? 

The value is derived from the configuration, where we are setting the latest version in the config.toml

https://github.com/keptn/keptn.github.io/blob/c43b29bd99f9b56f126fe650c4701aeca4605f25/config.toml#L22-L24

this means, as soon as we set this version for a newer one, for eg. our SEO optimizations, we also automatically change those links to the newest version documentation

PR #1574, opened 2023-01-18
Fix sitemap last modified date

aepfli:
If we do not find a proper date for a page, we assume the build date.

### DIFF
--- a/layouts/sitemap.xml
+++ b/layouts/sitemap.xml
@@ -2,7 +2,7 @@
 {{ $versionUsed := print "docs/" $.Site.Params.version}}
 {{ range .Data.Pages -}}
     {{ if not .Params.private -}}
-        {{ if or (in .Permalink $versionUsed) (not (in .Permalink "docs/"))}}
+        {{ if or (in .RelPermalink $versionUsed) (not (in .RelPermalink "docs/"))}}
             {{ $lastModDate := .Lastmod -}}
             {{ if not .Lastmod -}}
                 {{ $lastModDate = now -}}
--- a/layouts/sitemap.xml
+++ b/layouts/sitemap.xml
@@ -3,10 +3,7 @@
 {{ range .Data.Pages -}}
     {{ if not .Params.private -}}
         {{ if or (in .RelPermalink $versionUsed) (not (in .RelPermalink "docs/"))}}
-            {{ $lastModDate := .Lastmod -}}
-            {{ if not .Lastmod -}}
-                {{ $lastModDate = now -}}
-            {{- end}}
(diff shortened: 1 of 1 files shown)


layouts/sitemap.xml
```diff
@@ -2,14 +2,18 @@
 {{ $versionUsed := print "docs/" $.Site.Params.version}}
 {{ range .Data.Pages -}}
     {{ if not .Params.private -}}
-        {{ if or (in .Permalink $versionUsed) (not (in .Permalink "docs"))}}
+        {{ if or (in .RelPermalink $versionUsed) (not (in .RelPermalink "docs/"))}}
+            {{ $lastModDate := .Lastmod -}}
+            {{ if not .Lastmod -}}
+                {{ $lastModDate = now -}}
+            {{- end}}
```
mowies, line 9: why not put that into an if/else instead of first setting and then checking?

aepfli: I have yet to try to determine if this assumption is valid. Still, I usually need to define variables outside the condition anyways, so in the end, I favor just checking the error case and handling it, as it is fewer lines, and from a logic perspective, I think it is easier to follow

aepfli: i also coud do somethink like following based on https://gohugo.io/functions/cond/ which is actually a pretty nice and sweet idea
```suggestion
            {{ cond .Lastmod .Lastmod  now }}
```

aepfli: > i also coud do somethink like following based on https://gohugo.io/functions/cond/ which is actually a pretty nice and sweet idea

or
```suggestion
            {{ $lastModDate := default .Lastmod now }}
```

checking my options

mowies: i would prefer the `default` oneliner :)

layouts/sitemap.xml
```diff
@@ -2,14 +2,15 @@
 {{ $versionUsed := print "docs/" $.Site.Params.version}}
 {{ range .Data.Pages -}}
     {{ if not .Params.private -}}
-        {{ if or (in .Permalink $versionUsed) (not (in .Permalink "docs"))}}
+        {{ if or (in .RelPermalink $versionUsed) (not (in .RelPermalink "docs/"))}}
+        {{ $lastModDate := default now .Lastmod}}
```
mowies, line 6:
```suggestion
            {{ $lastModDate := default now .Lastmod}}
```

aepfli: thank you resolved

mowies (approved):

PR #1580, opened 2023-01-24
Folder and Version difference hotfix

aepfli:
Currently we do have version 1.1.x but our folders are still 1.0.x - we use the same variable for version within the docs and for our SEO improvements. By raising the version, and not renaming also the directory all our canonical urls broke, and even our sitemap was not fully populated. To mitigate this, I introduced another variable, called `versionDirectory` which states, where our documentation is currently sitting and adapted all the necessary snippets.

### DIFF
--- a/content/404/_index.md
+++ b/content/404/_index.md
@@ -1,3 +1,4 @@
 ---
 title: Page Not Found
+private: true
 ---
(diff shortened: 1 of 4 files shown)


content/404/_index.md
```diff
@@ -1,3 +1,4 @@
 ---
 title: Page Not Found
+private: true
```
aepfli, line 3: to  hide the page within the sitemap.xml

thisthat (approved):

Issue #1594, opened 2023-02-08
Addition of prerequisite for helm in website docs


SuperAayush:
Under the installation process in docs, the section of [Choose access options](https://keptn.sh/docs/install/access/) does contain the helm repo add information but directly contains the next step of helm charts installation.

A prerequisite can be added to add the helm keptn repo first before going for helm installation as discussed in today's community meeting.

The command could be: helm repo add keptn https://charts.keptn.sh

For more information visit here: https://artifacthub.io/packages/helm/keptn/keptn

cc- @DavidPHirsch @agardnerIT @bradmccoydev

SuperAayush:
I would like to work on this!!

SuperAayush:
Before starting to make changes I wanted to gather some thoughts regarding these two pages:
1. [Choose access options](http://localhost:1313/docs/1.0.x/install/access/)
2. [Install Keptn using the Helm chart](http://localhost:1313/docs/1.0.x/install/helm-install/)

I feel that these two pages contain pretty much the same information and content and can be merged to make a single page 🤔. Please let me know your thoughts on this and also if something else can be done.
OR
We can maybe swap the two pages which can also be a solution.

cc- @agardnerIT @DavidPHirsch

DavidPHirsch:
I would be in favor of merging

agardnerIT:
I agree that both these pages are conceptually very close to one another and can probably be merged.

ashutosh887:
@SuperAayush 
If merged, let's close this issue 😄

SuperAayush:
Hey @ashutosh887!!

There is still some work left on this issue, I am in the middle of my exams and would work on it after they are finished.
I still have some changes to make as per the suggestion of Meg and Adam.
Thank you for giving me a reminder.

ashutosh887:
Yeah please take your time @SuperAayush 
Best of luck to you!

StackScribe:
Keptn V1 has reached EOL on December 22nd, 2023, so this issue has been closed.

PR #1609, opened 2023-02-23
docs: structured and added helm prerequisite

SuperAayush:
As per the discussion in this PR, I have made some changes in the docs referring to the installation of Keptn.
Few things I did:

1. Merged the two pages into one - [Choose access options](https://keptn.sh/docs/install/access/) and [Install Keptn using the Helm chart](https://keptn.sh/docs/install/helm-install/).
2. Named the new page as **Access options and Installation of Keptn using Helm chart**.
3. Deleted **Install Keptn using the Helm chart** page.
4. Changed the weights of the pages as per the structure mentioned in [here](https://github.com/keptn/keptn.github.io/blob/master/CONTRIBUTING.md).
5. Made some typo fixes in CONTRIBUTING.md.

closes - #1594 

cc - @agardnerIT @StackScribe @DavidPHirsch


agardnerIT:
Thanks for the PR! A few thoughts and questions:

This PR seems to do a few things (updating typos and changing page weights). In future, please raise multiple seperate PRs rather than addressing too many "problems" in one PR.

Why?
1) It makes things clearer for the reviewer who, by definition and design, is further from your work and original intention that you.
2) The changelog and release notes are auto-generated from the PR title so subtle changes **inside** a PR can be lost (ie. In this case, in 6 months, someone looking at the docs release notes would have no way of quickly searching to see why the page weights were changed).

That aside:

1. Why are the page weights changed. I have no idea what page weighting does so perhaps it is completely logical, I just don't know what those numbers even mean.
2. Having said the above, I'm fine with the typo fix being in the contributors file - it's such a small change, the overhead of an additional PR would be a ridiculous ask :)

The only thing I'd like addressed before this is merged:

The old (now deleted in this PR) `helm/install` page includes a section describing the `logical` (but not physical) different between the "control plane" and the "execution plane". IMO this should not be lost as it is a frequent confusion point for users and so should be moved to the new page.

--------

I'm being verbose above to describe my thoughts, please do not take that as criticism. An excellent PR, well done!

SuperAayush:
Will never take any of your constructive suggestions as criticism 😄.
Paying focus to your suggestions:

1. Did not go for the multi PRs since the typos were very small and I was worried I would forget to create an issue for them. Really sorry for that and will never do it again!!
2. Will definitely work on making separate informative PRs for subtasks from now on!!
3.  What does weight number mean and why are they changed:
The weight number in the docs describes the file structure.
For example, in this PR I have changed the weight in order to maintain the continuity of the installation process.
weight
If you see the above screenshot the files **Create or bring a Kubernetes cluster**, **Kubernetes support & Cluster size**, and so on in the installation process are all weighted with a number to be in order. Similarly the file **Kubernetes support & Cluster size** has a greater weight than **Create or bring a Kubernetes cluster** file i.e 20 and 10. But If I reverse the weight the position of the file in the installation process might change which can be fatal for Keptn installation!!
For example, in the below screenshot, I reversed the weight number and as result, the positioning of the files also changed:
Weight1
4. I understand your concern about the "control plane" and the "execution plane" differentiation, but I have not removed that and only merged the content of `helm/install` into the access options page.
Merge

As an alternative, I can make a new page and place the difference in there for folks to pay major focus on that page.



I hope this message makes sense and looking forward to learning more good practices of contributions from you 😄.

cc- @agardnerIT

agardnerIT:
Any references to version `0.19.3` should be updated to `1.2.0` (@StackScribe is there a way to use a variable and make this dynamic so we don't have to continuously update the version numbers?)

I still can't see the control plane vs. exec. plane section - I'm more than happy to admit I'm blind though!

I'm not actually an approver so we'll need to wait for someone else to approve :)

StackScribe:
Note that these modifications are being made to the 0.19.x version of the Installation documentation.  The 1.y.z version of Installation instructions is under the 1.y.z header.  Yes, this is a problem -- a long story about how we got here and I haven't figured out an easy answer.  But it makes sense to get this rewrite set for one version and then apply the changes to the other version.

StackScribe:
Given that this is the 0.19.x version of the doc, the use of the 0.19.3 string for commands is correct.  @agardnerIT , the files are coded to automatically pick up the current version number when that is wanted, like in sample command lines.  This is explained in the CONTRIBUTING.md file but basically the current version is defined in the config.toml file and then we use a param string to provide the version number.  Note that we do not want to do this globally because we want strings like "Upgrading from release 0.18.x to 0.19.x" to be left alone.

StackScribe:
These two pages do need to be combined so thanks for taking this project on.  However, I think you need to do a bit more rewriting/combining the information than you have to get a good narrative flow.  I start reading about the access options and those include instructions for installing Keptn with each option, but you haven't yet told me the basics about how to install Keptn.

I think you need to start this page with the first few paragraphs under "Install Keptn using the Helm Chart".  And then maybe explain what the access options are and put in the big graphic, then have the specific instructions for each access option underneath that.

This title is really long and I think "Install using Helm Chart" should be the prominent part of the title.  I'm a little bit inclined to remove the "access options" from the title, unless that confuses users who are looking for the old page.  And be sure to check xrefs -- we have a LOT of xrefs to both of the old pages and they need to be modified.  I'm not sure why the xref checker didn't run against this PR -- that will at least help you find the broken xrefs.

I did some work on these two pages and I think some of that got lost -- we did have some file problems that would explain it.  So also note that we have some sample command lines that are not displaying as code blocks but rather as a long black fox then followed by the command (shown in normal font).  Also, I see at least one instance of "Setup" used as a verb when the verb form should be "Set up".  Sorry about that -- holler if you need help but, as I recall, the fix was pretty obvious when I looked at the source.

Again, kudos for taking on this bit of writing.  It really does need to be done!  I'm actually thinking that you should work on the 1.0.x version of the files first since that is the current version.  Then, when you get that set, if you want to backport the changes to 0.19.x, you can.

StackScribe:
Keptn V1 has reached EOL on December 22nd, 2023, so this PR has been closed.

PR #1621, opened 2023-03-14
log levels for debugging projects

StackScribe:
Preserve information from Moritz and Florian in https://keptn.slack.com/archives/CNRCGFU3U/p1678736139343539 .

### DIFF
--- a/content/docs/1.0.x/manage/debug/index.md
+++ b/content/docs/1.0.x/manage/debug/index.md
@@ -1,0 +1,24 @@
+---
+title: Debug logging
+description: Implement debug logging for your project
+weight: 45
+keywords: [1.0.x-manage]
+aliases:
+---
+
+Modifying the log level preserves more detailed information in the logs.
+This can be useful when analyzing execution problems for your project.
+
+To do this:
+
+* Adjust the value of the `LOG_LEVEL` environment variable
+  in the shipyard controller to `debug`.
+  Alternately, you can adjust the `loglevel` Helm value to `debug`.
+
+* Also adjust the value of the `LOG_LEVEL` environment variable
+  for the `resource-service` being used
+  at the time of the failure you are analyzing.
+
+Remember to reset the `LOG_LEVEL` environment variables
+back to `info` when you have finished your debugging effort.
+


mowies (changes requested):
Every service in keptn uses those log levels. So the `LOG_LEVEL` env variable exists in every keptn service to set that service's log level.
The `logLevel` helm value on the other hand is a global setting that sets the log level for **all** core services of keptn.
I think we should clarify that in the doc.

StackScribe:
Thanks for the suggestions -- I think this is clearer.

Should we also mention what happens if the helm value and environment variable do not match -- which takes precedence?

content/docs/1.0.x/manage/debug/index.md
```diff
@@ -9,14 +9,20 @@ aliases:
 Modifying the log level preserves more detailed information in the logs.
 This can be useful when analyzing execution problems for your project.
 
-To do this:
+You can set the log level in either of two ways:
 
-* Adjust the value of the `LOG_LEVEL` environment variable
-  in the shipyard controller to `debug`.
-  Alternately, you can adjust the `loglevel` Helm value to `debug`.
+* Adjust the `loglevel` value to `debug`
```
mowies, line 14: The right capitalization is important here
```suggestion
* Adjust the `logLevel` helm value to `debug`
```

content/docs/1.0.x/manage/debug/index.md
```diff
@@ -9,14 +9,20 @@ aliases:
...
+  This is a global setting that sets the log level
+  for **all** Keptn core services.
 
-* Also adjust the value of the `LOG_LEVEL` environment variable
-  for the `resource-service` being used
+* All Keptn services support the `LOG_LEVEL` environment variable.
+  Set this to `debug` to set the log level for an individual service.
+  In most cases, you will get the debugging information you need
+  by setting this for the `shippy` shipyard controller
+  and for the `resource-service` being used
```
mowies, line 25:
```suggestion
  by setting this for the shipyard-controller
  and/or for the `resource-service` being used
```

content/docs/1.0.x/manage/debug/index.md
```diff
@@ -9,14 +9,20 @@ aliases:
...
-To do this:
+You can set the log level in either of two ways:
 
-* Adjust the value of the `LOG_LEVEL` environment variable
-  in the shipyard controller to `debug`.
-  Alternately, you can adjust the `loglevel` Helm value to `debug`.
+* Adjust the `loglevel` value to `debug`
+  in the [Helm chart](../../reference/files/values)
+  or by using the [-set flag](../../install/helm-install/#the---set-flag)
+  on the **helm** command line.
```
mowies, line 17: Using `--set` to set values is generally not the best practise way of setting value, I suggest not advertise it. Generally, we have docs on how to set helm values anyways, I don't think we need to reiterate them here.
```suggestion
  in the [Helm chart](../../reference/files/values).
```

StackScribe: Not reiterating but xref'ing to the other info.  But it's gone.

content/docs/1.0.x/manage/debug/index.md
StackScribe: Done

content/docs/1.0.x/manage/debug/index.md
StackScribe: Done

mowies (approved):

PR #1633, opened 2023-04-20
Docs/add klt homepage link

agardnerIT:
This PR:

- Informs k8s native users that they are looking for the KLT and provides a link to the sub site.

### DIFF
--- a/content/home/intro.md
+++ b/content/home/intro.md
@@ -5,3 +5,9 @@
 - Operations & remediation
 
 declarative, extensible and based on GitOps
+
+### Cloud Native?
+
+You need the [Keptn Lifecycle Toolkit](https://lifecycle.keptn.sh)
+
+### Non Cloud Native?
--- a/content/home/intro.md
+++ b/content/home/intro.md
@@ -8,6 +8,6 @@ declarative, extensible and based on GitOps
 
 ### Cloud Native?
 
-You need the [Keptn Lifecycle Toolkit](https://lifecycle.keptn.sh)
+You need the <a href="https://lifecycle.keptn.sh" target="_blank" style="color: #fff; text-decoration: underline">Keptn Lifecycle Toolkit</a>.
 
 ### Non Cloud Native?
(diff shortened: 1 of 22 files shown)


agardnerIT:
@StackScribe @aepfli I don't get what I expect in this PR. I expect the words `Keptn Lifecycle Toolkit` to link to the sub site. However in the browser dev console I see `<!-- raw html omitted -->`... I need to use raw HTML as I need to custom style the link.

If I use plain markdown, the link is coloured blue and is un-distinguishable from the background.

@thisthat as we discussed today.

content/home/intro.md
```diff
@@ -4,4 +4,10 @@
 - SLO-driven multi-stage delivery 
 - Operations & remediation
 
-declarative, extensible and based on GitOps
\ No newline at end of file
+declarative, extensible and based on GitOps
+
+### Cloud Native?
```
aepfli, line 9: i think we also need to adapt our catch phrase then

image

because it might offer some confusion

aepfli:
should it looks like this? 
image

or is the button wrong?

content/home/intro.md
agardnerIT: > should it looks like this? 
> or is the button wrong?

Personally I like the buttons, they draw attention.

For the slogan, perhaps remove "Cloud native" from the front? So it's just "application lifecycle orchestration" then a user can pick cloud native or not?

Not sure if it is "Cloud-native" or "Cloud native". Most recent signage I've seen has it as "cloud native". Same for "life-cycle". I think it may be "lifecycle"? [Merriam Webster has it as "life cycle"](https://www.merriam-webster.com/dictionary/life%20cycle) so I'm leaning towards the separate versions

aepfli:
i created https://github.com/keptn/keptn.github.io/pull/1634 which should actually do what you want it to be doing (not ideal and hacky) but for a fast solution it should be great

aepfli:
superseeded by https://github.com/keptn/keptn.github.io/pull/1634

PR #1646, opened 2023-06-01
docs: update slack link

staceypotter:
change from #advocacy-and-outreach to #community-and-outreach

### DIFF
--- a/content/community/_index.md
+++ b/content/community/_index.md
@@ -6,7 +6,7 @@ weight: 20
 
 👋 **Welcome to the Keptn community!**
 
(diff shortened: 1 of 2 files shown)


rakshitgondwal:
Following up on @agardnerIT's comment: I agree with the point that this had nothing to do with the PR but we were able to point out a potential issue at the end, so yes this issue was not related to the PR but it had to be fixed one or other day so the linter is correct on its side in my opinion. Would love to hear your thoughts too.

StackScribe (approved):
Nice job!

StackScribe:
As for the general issue about the bad link to the Code of Conduct...  We are getting these all over the place.  I'm trying to figure out why the PR that moved that file didn't pick up the xrefs that were broken so they could be fixed right then.

PR #1655, opened 2023-06-20
docs:  updated the supported versions table 

Nitishupkr:
Fixes: #1656


StackScribe (commented):
You need to leave the info for 0.18.x and 0.19.x in place -- we still have a lot of users on those versions.  You need to add a new row and put the info for 1.* in the new row.

Nitishupkr:
> You need to leave the info for 0.18.x and 0.19.x in place -- we still have a lot of users on those versions. You need to add a new row and put the info for 1.* in the new row.

solved this check it.

StackScribe:
Keptn V1 has reached EOL on December 22nd, 2023, so this PR has been closed.

